PR #3, opened 2018-04-27
Add real node pool support

mikkeloscar:
This adds node pool support as discussed internally. (For any external interested parties we will document better how this will work, don't worry :)).

The code in `clusterpy.go` and `aws.go` is a bit ugly because it's supporting both the current design and the new node pool design at the same time. I have put a bunch of `if nodePoolFeatureEnabled(cluster)` guards in the code which enables the the node pool feature if the config item `node_pool_feature_enabled=true` is set in the cluster registry. All of this can be removed once all clusters uses the new node pool feature.

### The way it will work:

1. Set the `node_pool_feature_enabled=true` config item on a cluster that should be migrated.
2. Add two extra node pools to the cluster `default-master` & `default-worker`.
3. The CLM will create the extra node pools and gracefully scale down the old node pools to 0 (while still keeping them in the `senza-definition.yaml`.
4. Once all clusters has been migrated we can remove the ASG definitions from `senza-definition.yaml` and remove the featuregate code from CLM.

This depends on these changes in kubernetes-on-aws: https://github.com/zalando-incubator/kubernetes-on-aws/pull/1005

### Naming

Since naming is always hard I want to make sure it's not totally bad as it's a bit annoying to change later on. I have defined the following names for tags/labels etc.:

* Tags on node pools
  * `kubernetes.io/node-pool=<name>` Name of the node pool (will also be a label on each Kubernetes node).
  * `kubernetes.io/node-pool/profile=<profile>` Profile name of the node pool (from cluster registry).
  * `kubernetes.io/role/node-pool=true` Tag that can be used for filtering all node pool stacks/instances belonging to *some* node pool.
* Node pool stack name format: `nodepool-<nodepool-name>-<clusterid>`


szuecs:
Is this a wip or ready to review?

mikkeloscar:
@szuecs it's ready for review :)

pkg/updatestrategy/aws_asg.go
```diff
@@ -129,23 +131,128 @@ func (n *ASGNodePoolsBackend) Scale(nodePool *api.NodePool, replicas int) error
...
+		resp, err := n.asgClient.DescribeAutoScalingGroups(asgParams)
+		if err != nil {
+			return err
+		}
+
+		if len(resp.AutoScalingGroups) == 0 {
+			return fmt.Errorf("failed to find ASG '%s'", asgName)
+		}
+
+		asg := resp.AutoScalingGroups[0]
```
szuecs: I guess this is not working with multiple ASGs, right?

pkg/updatestrategy/aws_asg.go
```diff
@@ -129,23 +131,128 @@ func (n *ASGNodePoolsBackend) Scale(nodePool *api.NodePool, replicas int) error
...
+
+		newDesired := aws.Int64Value(asg.DesiredCapacity) - 1
+		minSize := aws.Int64Value(asg.MinSize)
+
+		if minSize > 0 && newDesired >= 0 && minSize > newDesired {
```
szuecs: nit: I think it would be better readable if you sort these like this, because you have at left most the smalles and rightmost the largest:

    if 0 <= newDesired && 0 < minSize && newDesired < minSize {

pkg/updatestrategy/aws_asg.go
mikkeloscar: Yes, this will only work if a node pool is a single ASG. I have some patches for mapping a single node pool to multiple ASGs, but I don't want to include it here as the change becomes too big IMO. So that would be a follow-up after this is merged and rolled out.

provisioner/aws.go
```diff
@@ -203,24 +206,65 @@ func (a *awsAdapter) CreateOrUpdateClusterStack(stackName, stackDefinitionPath s
...
+			workerPoolDesired = aws.Int64Value(asg.DesiredCapacity)
+			workerPoolMinSize = aws.Int64Value(asg.MinSize)
+			workerPoolMaxSize = workerPoolDesired
+
+			// get master nodepool size
+			asg, err := a.getNodePoolASG(cluster.ID, masterPool.Name)
+			if err != nil {
+				if err != ErrASGNotFound {
+					return nil, err
+				}
```
szuecs: if err = ErrASGNotFound we create a worker ASG with 0 desired, does it makes sense?

provisioner/node_pools.go
```diff
@@ -0,0 +1,357 @@
...
+	// create ignition config pulling from s3
+	ignCfg = []byte(fmt.Sprintf(ignitionBaseTemplate, uri))
+
+	return base64.StdEncoding.EncodeToString(ignCfg), nil
+}
+
+// uploadUserDataToS3 uploads the provided userData to the specified S3 bucket.
+// The S3 object will be named by the sha512 hash of the data.
+func (n *AWSNodePoolProvisioner) uploadUserDataToS3(userData []byte, bucketName string) (string, error) {
+	// sha1 hash the userData to use as object name
```
szuecs: It's not sha1 -> drop "sha1"

provisioner/node_pools.go
```diff
@@ -0,0 +1,357 @@
...
+
+// NodePoolProvisioner is able to provision node pools for a cluster.
+type NodePoolProvisioner interface {
+	Provision(values map[string]string) error
+	Reconcile() error
+}
+
+// AWSNodePoolProvisioner is a node provisioner able to provision node pools
+// in AWS via cloudformation.
+type AWSNodePoolProvisioner struct {
```
szuecs: Based on the request for baremetal support, we should have a TODO here to refactor these types and methods into providers.

provisioner/node_pools.go
```diff
@@ -0,0 +1,357 @@
...
+	Values   map[string]string
+}
+
+type userDataParams struct {
+	Cluster  *api.Cluster
+	NodePool *api.NodePool
+	Values   map[string]string
+}
+
+func (n *AWSNodePoolProvisioner) generateNodePoolStackTemplate(nodePool *api.NodePool, values map[string]string) (string, error) {
```
szuecs: I would prefer `(npp *AWSNodePoolProvisioner)` or `*p`

provisioner/node_pools.go
mikkeloscar: Done.

provisioner/node_pools.go
mikkeloscar: Added a TODO.

provisioner/node_pools.go
mikkeloscar: Done (changed to `p`)

provisioner/aws.go
mikkeloscar: This is intended for the legacy node pool migration. I added a comment to make it more clear.

szuecs:
:+1:

PR #7, opened 2018-05-07
Switch from glide to dep

linki:
imo, it makes managing the dependencies a little easier than glide by now.


linki:
closed in favour of https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/8

PR #18, opened 2018-05-16
Drop unsupported clusters early on

aermakov-zalando:
Unsupported clusters should be completely ignored, otherwise CLM goes into an update loop (refresh -> find invalid status -> try to update -> bail out).


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #20, opened 2018-05-16
remove update command, add ARG instruction, change maintainer to label

ben-st:
i changed maintainer to label since former is deprecated as of docker version 1.13
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

i added a variable for the kubernetes version 

i changed the apk update to `apk add --no-cache` since  IMHO you only want to install packages and not update your system with e.g. `apk upgrade`.

All the best,

Benjamin

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,15 @@
 FROM registry.opensource.zalan.do/stups/alpine:latest
-MAINTAINER Team Teapot @ Zalando SE <team-teapot@zalando.de>
+LABEL maintainer="Team Teapot @ Zalando SE <team-teapot@zalando.de>"
+
+ARG K8S_VERSION=v1.9.5
 
 # install cluster.py dependencies
 # including kubectl
-RUN apk update && \
-    apk add python3 ca-certificates openssl git openssh-client && \
+RUN apk add --no-cache python3 ca-certificates openssl git openssh-client && \
     python3 -m ensurepip && \
     rm -r /usr/lib/python*/ensurepip && \
     pip3 install --upgrade stups-senza && \
-    wget -O /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.9.5/bin/linux/amd64/kubectl && \
+    wget -O /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$K8S_VERSION/bin/linux/amd64/kubectl && \
     chmod 755 /usr/local/bin/kubectl && \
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

szuecs:
thanks

Dockerfile
```diff
@@ -1,14 +1,15 @@
...
-MAINTAINER Team Teapot @ Zalando SE <team-teapot@zalando.de>
+LABEL maintainer="Team Teapot @ Zalando SE <team-teapot@zalando.de>"
+
+ARG K8S_VERSION=v1.9.5
 
 # install cluster.py dependencies
 # including kubectl
-RUN apk update && \
-    apk add python3 ca-certificates openssl git openssh-client && \
+RUN apk add --no-cache python3 ca-certificates openssl git openssh-client && \
```
mikkeloscar: I see your point about not updating but it's a trade-off between being stable vs. having the latest fixes/security updates (which could break). So far we have not had any problem with this, so I would prefer to leave it as it was.

The other changes are great! Thanks a lot!

mikkeloscar: Sorry, didn't realize `--no-cache` does the same thing.

mikkeloscar:
@ben-st please sign-off the commit to make the bot happy :)

ben-st:
@mikkeloscar the bot is expecting my noreply github email. Is this expectation correct or is it ok with this configuration as my email matches my git config. otherwise i will push the requested config the bot needs, but it looks strange to me in the first place.

mikkeloscar:
@ben-st hmm, it's because of this:

```
$ git show                                                                                            [0]
commit e9b3774607eb2c0519e6bdc1b664f7df053a25ae (HEAD -> ben-st-patch-1, ben-st/patch-1)
Author: Benjamin <ben-st@users.noreply.github.com>
Date:   Wed May 16 15:58:31 2018 +0200

    remove update command, add ARG instruction, change maintainer to label
    
    Signed-off-by: Benjamin Stein <info@diffus.org>
```

Did you create it the commit via the Github interface?

I'm sorry about this, we introduced the bot recently to enforce DCO and test how much of annoyance it would be.

You can change the author of your commit to fix this. See: https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

szuecs:
:+1:

hjacobs:
@mikkeloscar I think we have to remove "CDP build" from the required status checks as it will not be triggered for external contributors.

mikkeloscar:
:+1:

PR #21, opened 2018-05-17
git: support SHA versions, add tests

aermakov-zalando:
- Add support for SHA checkouts for git config source (these are used in tests)
- Add tests for both checkout modes


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #23, opened 2018-05-17
Support multi ASG node pools in update strategy

mikkeloscar:
This adds support for node pools that span multiple Auto Scaling Groups in the update strategy code. This is one part of enabling support for node pools where there is one ASG per Availability Zone which should make cluster auto-scaling much more reliable.

Note: This does not implement provisioning of node pools that span multiple ASGs, that should be done separately.

It's fully backwards compatible with node pools that only span a single ASG.

### DIFF
--- a/pkg/updatestrategy/aws_asg.go
+++ b/pkg/updatestrategy/aws_asg.go
@@ -3,6 +3,7 @@ package updatestrategy
 import (
 	"fmt"
 	"math"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/updatestrategy/aws_asg.go
```diff
@@ -58,77 +59,138 @@ func NewASGNodePoolsBackend(clusterID string, sess *session.Session) *ASGNodePoo
...
+	}
 
-	params := &autoscaling.UpdateAutoScalingGroupInput{
-		AutoScalingGroupName: asg.AutoScalingGroupName,
-		DesiredCapacity:      aws.Int64(int64(replicas)),
-		MinSize:              aws.Int64(min),
-		MaxSize:              aws.Int64(max),
+	diff := replicas - desired
+
+	if diff == 0 {
```
szuecs: <= would be more safe

szuecs:
:+1:

mikkeloscar:
:+1:

PR #24, opened 2018-05-18
Remove Outputs field from Cluster struct (was a hack)

mikkeloscar:
Removes a hack that was put in place when cloudformation didn't support
static naming of ASG lifecycle hooks. Now we don't need it anymore.

The need for this was removed with: https://github.com/zalando-incubator/kubernetes-on-aws/pull/991


szuecs:
:+1:

mikkeloscar:
:+1:

PR #25, opened 2018-05-18
Allow updates if we don't depend on any clusters

aermakov-zalando:
We have one channel that only has a production cluster, and currently it's blocked.


mikkeloscar:
:+1:

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #26, opened 2018-05-18
[WIP] Support blocking updates

aermakov-zalando:
Add support for blocking cluster updates. If a config item (`cluster-update-block`) is present, the cluster will not be selected for updates. Additionally, existing updates will be cancelled as fast as possible.


api/cluster.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/zalando-incubator/cluster-lifecycle-manager/channel"
 )
 
+const (
+	updateBlockedConfigItem = "cluster-update-block"
```
mikkeloscar: I think we enforce underscores instead of hyphens in Cluster registry.

provisioner/clusterpy.go
```diff
@@ -212,10 +237,14 @@ func (p *clusterpyProvisioner) Provision(cluster *api.Cluster, channelConfig *ch
 
 			sort.Sort(api.NodePools(nodePools))
 			for _, nodePool := range nodePools {
-				err := updater.Update(context.Background(), nodePool)
+				err := updater.Update(ctx, nodePool)
 				if err != nil {
 					return err
 				}
+
+				if err = ctx.Err(); err != nil {
```
linki: Isn't this superflous? When the context was cancelled wouldn't `updater.Update(ctx, nodePool)` right above already return a non-nil `err`?

aermakov-zalando: True, but the performance of the check is rather negligible and it's easier to just keep them between all major steps on every level.

provisioner/clusterpy.go
```diff
@@ -334,15 +367,18 @@ func (p *clusterpyProvisioner) Decommission(cluster *api.Cluster, channelConfig
 		logger.Error("Unable to downscale the deployments, proceeding anyway: %s", err)
 	}
 
+	// we don't support cancelling decommission operations yet
```
linki: Why not? Doesn't this violate the principle of least surprise? It should just work as CLM can already die at any moment.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #27, opened 2018-05-21
Wait for termination grace period when evicting

mikkeloscar:
This solves two related issues. The first issue is that we always
evicted with graceperiod 0 meaning the Pod Disruption Budget would be
respected, but pods would be instantly killed when evicted, not getting
the promised grace period to shutdown.

Even when the graceperiod is respected we could still end up in a
situation where a pod would be evicted and then the node would be
terminated right after, again not allowing pods to gracefully terminate.
This is solved by waiting grace period + (some buffer time) after the
eviction has been triggered, giving the pods the promised graceperiod to
shut down plus an extra 30 sec (same as the cluster-autoscaler uses) to
terminate gracefully.

### DIFF
--- a/pkg/updatestrategy/node_pool_manager.go
+++ b/pkg/updatestrategy/node_pool_manager.go
@@ -410,28 +410,27 @@ var evictPod = func(client kubernetes.Interface, logger *log.Entry, pod *v1.Pod)
 // This is to fully respect the termination expectations as described in:
 // https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods
 func waitForPodTermination(client kubernetes.Interface, pod v1.Pod) error {
-	if pod.Spec.TerminationGracePeriodSeconds != nil {
-		waitForTermination := func() error {
-			_, err := client.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
-			if err != nil {
-				if apiErrors.IsNotFound(err) {
-					return nil
-				}
-				return err
-			}
-			return fmt.Errorf("pod not terminated")
-		}
+	if pod.Spec.TerminationGracePeriodSeconds == nil {
+		// if no grace period is defined, we don't wait.
+		return nil
+	}
 
-		gracePeriod := time.Duration(*pod.Spec.TerminationGracePeriodSeconds)*time.Second + podEvictionHeadroom
-
-		backoffCfg := backoff.NewExponentialBackOff()
-		backoffCfg.MaxElapsedTime = gracePeriod
-		err := backoff.Retry(waitForTermination, backoffCfg)
+	waitForTermination := func() error {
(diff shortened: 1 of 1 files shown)


pkg/updatestrategy/node_pool_manager.go
```diff
@@ -368,17 +380,57 @@ var evictPod = func(client kubernetes.Interface, logger *log.Entry, pod *v1.Pod)
...
+}
+
+// waitForPodTermination waits for a pod to be terminated by looking up the pod
+// in the API server.
+// It waits for up to TerminationGracePeriodSeconds as specified on the pod +
+// an additional eviction head room.
+// This is to fully respect the termination expectations as described in:
+// https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods
+func waitForPodTermination(client kubernetes.Interface, pod v1.Pod) error {
+	if pod.Spec.TerminationGracePeriodSeconds != nil {
```
linki: nit: I would rather change the if to check for `== nil` and then do an early return instead of having the entire method wrapped inan `if` statement.

linki:
lgtm

linki:
👍

pkg/updatestrategy/node_pool_manager.go
mikkeloscar: Fixed!

njuettner:
👍

mikkeloscar:
:+1:

aermakov-zalando:
This has the same issue as the cluster-autoscaler shutdown logic. Statefulset pods can be restarted between poll attempts, so you need to check whether the `uid` is the same as well.

pkg/updatestrategy/node_pool_manager.go
```diff
@@ -368,17 +380,57 @@ var evictPod = func(client kubernetes.Interface, logger *log.Entry, pod *v1.Pod)
...
 
 	err := client.CoreV1().Pods(pod.Namespace).Evict(eviction)
 	if err != nil {
 		return err
 	}
+	localLogger.Info("Evicting pod")
+
+	// wait for the pod to be actually evicted and gone from the node.
+	// It has TerminationGracePeriodSeconds time to clean up.
+	start := time.Now().UTC()
```
szuecs: you don't need UTC()

pkg/updatestrategy/node_pool_manager.go
```diff
@@ -368,17 +380,57 @@ var evictPod = func(client kubernetes.Interface, logger *log.Entry, pod *v1.Pod)
...
+	err = waitForPodTermination(client, *pod)
+	if err != nil {
+		localLogger.Warnf("Pod not terminated within grace period: %s", err)
+	}
+
+	localLogger.Infof("Pod evicted. (Observed termination period: %s)", time.Now().UTC().Sub(start))
```
szuecs: You don't need UTC here

PR #28, opened 2018-05-22
Fix graceful termination for statefulset pods

aermakov-zalando:
Handle statefulset pods correctly (they have the same name after restarting, so we need to check the uid as well).

### DIFF
--- a/pkg/updatestrategy/node_pool_manager.go
+++ b/pkg/updatestrategy/node_pool_manager.go
@@ -416,14 +416,20 @@ func waitForPodTermination(client kubernetes.Interface, pod v1.Pod) error {
 	}
 
 	waitForTermination := func() error {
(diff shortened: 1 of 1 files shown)


pkg/updatestrategy/node_pool_manager.go
```diff
@@ -416,14 +416,20 @@ func waitForPodTermination(client kubernetes.Interface, pod v1.Pod) error {
...
 			}
-			return err
+			return backoff.Permanent(err)
+		}
+
+		// statefulset pods have the same name after restart, check the uid as well
+		if newpod.ObjectMeta.UID == pod.ObjectMeta.UID {
+			return fmt.Errorf("pod not terminated")
+		} else {
+			return nil
```
mikkeloscar: nit: the else branch is not needed, just `return nil` from the function scope.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #34, opened 2018-05-28
Read configuration defaults from a file

aermakov-zalando:
Check if `cluster/config-defaults.yaml` exists as the first step during provisioning. If it's there, evaluate it as a Go template with a cluster definition stripped of config items, parse it as a YAML file that's expected to be a simple string->string map, and set all missing config items to the values from the parsed map.

This has multiple benefits:
 - the defaults are in one place, even if they're used in multiple templates (could even use this for e.g. specifying hyperkube image)
 - the template is a bit cleaner because we don't have the same stupid `if index ...` everywhere

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,34 @@ the `name` will be used. If none of them are defined, it's an error.
 
 `kind` must be one of the kinds defined in `kubectl get`.
 
+## Configuration defaults
+
+CLM will look for a `config-defaults.yaml` file in the cluster configuration
+directory. If the file exists, it will be evaluated as a Go template with all
+the usual CLM variables and functions available, and the resulting output will
+be parsed as a simple key-value map. CLM will use the contents of the file to
+populate the cluster's configuration items, taking care not to overwrite the
+existing ones.
+
+For example, you can use the defaults file to have different settings for
+production and test clusters, while keeping the manifests readable:
+
+* **config-defaults.yaml**:
+    ```yaml
+    {{ if eq .Environment "production"}}
+    autoscaling_buffer_pods: "3"
+    {{else}}
+    autoscaling_buffer_pods: "0"
+    {{end}}
+    ```
+
+* **manifests/example/example.yaml**:
+    ```yaml
+    …
+    spec:
+      replicas: {{.ConfigItems.autoscaling_buffer_pods}}
+    …
+    ```
 
 ## Non-disruptive rolling updates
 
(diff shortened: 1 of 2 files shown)


mikkeloscar:
Please document this in the `README.md` similar to how the deletions file is documented.

provisioner/clusterpy.go
```diff
@@ -97,6 +97,36 @@ func (p *clusterpyProvisioner) Supports(cluster *api.Cluster) bool {
 	return cluster.Provider == providerID
 }
 
+func (p *clusterpyProvisioner) updateDefaults(cluster *api.Cluster, channelConfig *channel.Config) error {
+	defaultsFile := path.Join(channelConfig.Path, "cluster", "config-defaults.yaml")
```
mikkeloscar: Please define a `const` for `config-defaults.yaml`

provisioner/clusterpy.go
```diff
@@ -97,6 +97,36 @@ func (p *clusterpyProvisioner) Supports(cluster *api.Cluster) bool {
...
+
+	var defaults map[string]string
+	err = yaml.Unmarshal([]byte(result), &defaults)
+	if err != nil {
+		return err
+	}
+
+	for k, v := range defaults {
+		_, exists := cluster.ConfigItems[k]
+		if !exists {
```
mikkeloscar: nit: it's common in go to use `ok` for this variable, just reads easier if it's the same name used. :)

mikkeloscar:
:+1:

aermakov-zalando:
👍

README.md
```diff
@@ -151,6 +151,34 @@ the `name` will be used. If none of them are defined, it's an error.
...
+For example, you can use the defaults file to have different settings for
+production and test clusters, while keeping the manifests readable:
+
+* **config-defaults.yaml**:
+    ```yaml
+    {{ if eq .Environment "production"}}
+    autoscaling_buffer_pods: "3"
+    {{else}}
+    autoscaling_buffer_pods: "0"
+    {{end}}
```
linki: What do you think about dedicated config files for each environment, like other projects do with `development.yaml`, `production.yaml` etc. In our case:`${cluster.Environment}.yaml`. That would lead to some duplication between the files but would get rid of the `if/else` logic in the config file.

aermakov-zalando: In my opinion a simple `if/else` is better than repeating the same default in multiple config files, but we could also have something like `config-defaults/global.yaml`. I would wait and see if we start abusing the templating, and then implement some better solutions. WDYT?

linki: Let's wait then. A `global.yaml` that the individual envs would "inherit" from would be overkill for now.

PR #35, opened 2018-05-29
[WIP] Evict pods in parallel

### LINKED ISSUE #30: Evict pods in parallel
Since #27 it now takes a "long time" (~10 min.) to evict pods from a single node even if it's not violating the PDB. This is because we evict the pods one by one and wait for each pod to be gracefully terminated.

There is no reason not to do this in parallel.

mikkeloscar:
**still WIP**

This makes the eviction of pods concurrent such that the CLM doesn't wait 3 days for each pod on a node, but instead only waits 3 days in total (worst case :)).

Fix #30


mikkeloscar:
This is now ready for review!

pkg/updatestrategy/node_pool_manager.go
```diff
@@ -278,97 +278,135 @@ func (m *KubernetesNodePoolManager) drain(ctx context.Context, node *Node) error
...
+	}
 
-		var lastPDBViolationErr error
-		for _, pod := range pods.Items {
-			// we check at the start because there's a continue in the loop body
-			if err = ctx.Err(); err != nil {
-				return backoff.Permanent(err)
-			}
+	errors := make(chan error, len(pods.Items))
+	for _, pod := range pods.Items {
```
aermakov-zalando: Can we extract it somewhere? The exact same logic is in node pool creation I think, at this point I think it makes sense to have something like `waitFor(funcs []func() error) []error`. Alternatively, we could just use `ErrGroup` and log the errors in the individual functions, which has the benefit of not having to implement anything.

mikkeloscar: Changed it to use `errgroup`.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #36, opened 2018-05-29
Extract templating

aermakov-zalando:
- Use the same templating code for manifests and node pool data
- Add basic tests
- Remove a useless os.Open

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -16,6 +16,7 @@ type applyContext struct {
 	manifestData          map[string]string
 	baseDir               string
 	computingManifestHash bool
+	readTemplate          func(string) ([]byte, error)
 }
 
 func newApplyContext(baseDir string) *applyContext {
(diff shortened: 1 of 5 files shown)


provisioner/template.go
```diff
@@ -0,0 +1,102 @@
...
+func newApplyContext(baseDir string) *applyContext {
+	return &applyContext{
+		baseDir:      baseDir,
+		manifestData: make(map[string]string),
+	}
+}
+
+// applyTemplate takes a fileName of a template and the model to apply to it.
+// returns the transformed template or an error if not successful
+func applyTemplate(context *applyContext, filePath string, data interface{}) (string, error) {
```
mikkeloscar: nit: I would rather call this something like `renderTemplate`. Apply, in my mind, is more in the context of `kubectl apply`.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #37, opened 2018-05-29
Add autoscalingBufferSettings template function

aermakov-zalando:
Given a cluster configuration, it returns the CPU and memory settings for the buffer pods using the following logic:

 - if `autoscaling_buffer_cpu` and `autoscaling_buffer_memory` are set, return their values directly
 - if only one of those is set, fail (this saves us from surprises if the node sizes change in the future)
 - find the largest instance from the pools matching the regex from `autoscaling_buffer_pools` config item, scale their CPU and memory by `autoscaling_buffer_cpu_scale` and `autoscaling_buffer_memory_scale` and return either these values, or  `instance_size - autoscaling_buffer_{cpu|memory}_reserved`, whichever is greater. Fail if the parameters are missing or invalid, or if    we can't determine the largest instance because some pools have more CPU and others more memory.

**Note**: this depends on #36, I'll rebase after it's merged.

### DIFF
--- a/pkg/aws/instance_info.go
+++ b/pkg/aws/instance_info.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"sync"
 
+	"fmt"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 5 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #41, opened 2018-05-31
Return stderr as part of command err

mikkeloscar:
In #38 we introduced a bug because `command.Run` no longer returned the stderr string as part of the `error`. This caused a problem in the `deletions.yaml` handling because we rely on the stderr output of `kubectl` to determine if the command just failed because the to-be-deleted resource just wasn't found.

This reverts the behavior.

https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/master/provisioner/clusterpy.go#L820-L825

### DIFF
--- a/pkg/util/command/command.go
+++ b/pkg/util/command/command.go
@@ -1,10 +1,11 @@
 package command
 
 import (
+	"fmt"
 	"os/exec"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
-	"strings"
 )
 
 func outputLines(raw []byte) []string {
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

njuettner:
👍

mikkeloscar:
:+1:

aermakov-zalando:
But this hides the output from the various `git` commands, since `git.go` doesn't print the output.

aermakov-zalando:
Actually, what benefit is there in running kubectl through the `util.Command` thing?

PR #42, opened 2018-05-31
util/command: Run->RunSilently, add Run

aermakov-zalando:
* Rename `Run` -> `RunSilently`
* `RunSilently`: don't print anything unless there's an error
* Add `Run`, which always prints the output (useful for e.g. kubectl)
* Return the combined output from both functions


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #45, opened 2018-05-31
Run tests with -race on Travis

aermakov-zalando:
We don't run CDP builds for external contributors, and `-race` could catch some errors.


aermakov-zalando:
`goveralls` doesn't support it, so it's a bit trickier to fix.

PR #51, opened 2018-06-05
Ignore terminated pods when evicting

aermakov-zalando:
PDBs are checked before the pod phase, so a broken PDB can prevent us from evicting terminated pods as well, even though it makes no sense. Just ignore them instead.


mikkeloscar:
:+1:

PR #52, opened 2018-06-07
Node pools: fix data sharing issues

aermakov-zalando:
* Create a private copy of Values for every node pool
* Change to `map[string]interface{}` to allow non-string values


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #53, opened 2018-06-08
Minor bugfixes

aermakov-zalando:
* Set `node-pool/profile` tag to the actual profile name, not node pool name
* Correctly print the notice about terminated pods


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #55, opened 2018-06-11
Check if the pod was already moved before evicting

### LINKED ISSUE #54: Re-fetch pod information before evicting
We have a race condition because the Kubernetes Eviction API doesn't properly support statefulset pods (https://github.com/kubernetes/kubernetes/issues/64923). Let's just re-fetch the pod metadata on every eviction attempt, this way we can at least reduce the race window to one or two seconds.

aermakov-zalando:
It's possible that someone else might move the pod while we're evicting it. Since statefulset pods have the same name, CLM will try to evict it again even if it's not running on the node we're trying to drain. Re-check on every attempt to reduce the race window. Fix #54.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #56, opened 2018-06-12
Add support for config items per node pool

mikkeloscar:
Adds support for config items per node pool.

(Also added some other cosmetic changes to the swagger definition to align it with our internal cluster-registry definition).


linki:
👍

PR #59, opened 2018-06-13
Limit surge to desired nodes during rolling update

mikkeloscar:
In case desired < surge we don't want to scale up by surge new nodes,
instead we limit it to desired. This ensures that a node pool with
desired 0 doesn't scale up during a rolling update.


aermakov-zalando:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #60, opened 2018-06-14
Re-check if the pod terminated on every attempt

aermakov-zalando:
A pod might finish while we're trying to evict it, so we need to check on every attempt.


aermakov-zalando:
👍

linki:
👍

PR #61, opened 2018-06-18
Wait for instance to be terminated in AWS

mikkeloscar:
Deleting the Kubernetes node object when terminating a node had the bad side effect that flannel on other nodes would remove the routes before the node was actually gone. Instead of deleting the node object we wait for the instance to be terminated according to AWS this way we don't drain the same node multiple times (which was the reason for deleting the node object manually in the first place).

### DIFF
--- a/pkg/updatestrategy/aws_asg.go
+++ b/pkg/updatestrategy/aws_asg.go
@@ -1,6 +1,7 @@
 package updatestrategy
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"sort"
@@ -15,6 +16,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/elb/elbiface"
+	"github.com/cenkalti/backoff"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
 )
 
@@ -236,10 +238,11 @@ func (n *ASGNodePoolsBackend) deleteTags(nodePool *api.NodePool, tags map[string
 	return err
 }
 
-// Terminate terminates a node from the ASG and optionally decrements the
+// Terminate terminates an instance from the ASG and optionally decrements the
 // DesiredCapacity. By default the desired capacity will not be decremented.
 // In case the new desired capacity is less then the current min size of the
 // ASG, it will also decrease the ASG minSize.
+// This function will not return until the instance has been terminated in AWS.
 func (n *ASGNodePoolsBackend) Terminate(node *Node, decrementDesired bool) error {
 	instanceId := instanceIDFromProviderID(node.ProviderID, node.FailureDomain)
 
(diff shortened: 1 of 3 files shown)


pkg/updatestrategy/aws_asg.go
```diff
@@ -316,26 +319,40 @@ func (n *ASGNodePoolsBackend) Terminate(node *Node, decrementDesired bool) error
 	}
 
 	_, err := n.asgClient.TerminateInstanceInAutoScalingGroup(params)
```
aermakov-zalando: We should still handle this `err` somehow. If it fails, debugging this would be very hard.

mikkeloscar: It's handled down here after checking the status first: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/61/files#diff-41f826bed623fac059b975e38370daffR348

aermakov-zalando: Right, I've missed that. It's very weird that we retry `instanceStatus` for potentially a very long time and then suddenly return an error that happened days ago in a completely different code, would you mind either refactoring it somehow or adding a comment?

mikkeloscar: Refactored it to be more clear. Now, on error, it will try to get the state (but not care about it) and only return an error if there's also an error on getting the current instance state.

Otherwise it will just go into the backoff waiting for the instance to be terminated.

pkg/updatestrategy/aws_asg.go
```diff
@@ -317,25 +321,52 @@ func (n *ASGNodePoolsBackend) Terminate(node *Node, decrementDesired bool) error
...
-		case ec2.InstanceStateNameShuttingDown, ec2.InstanceStateNameStopping, ec2.InstanceStateNameTerminated, ec2.InstanceStateNameStopped:
+		switch state {
+		case ec2.InstanceStateNameShuttingDown, ec2.InstanceStateNameStopping:
+			log.Info("shutting down")
+			return errors.New("instance shutting down")
+		case ec2.InstanceStateNameTerminated, ec2.InstanceStateNameStopped:
 			return nil
 		default:
-			return err
+			return nil
```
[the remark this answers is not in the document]
mikkeloscar: Fixed

pkg/updatestrategy/aws_asg.go
```diff
@@ -317,25 +320,51 @@ func (n *ASGNodePoolsBackend) Terminate(node *Node, decrementDesired bool) error
...
+		case ec2.InstanceStateNameTerminated, ec2.InstanceStateNameStopped:
 			return nil
 		default:
-			return err
+			return fmt.Errorf("unexpected instance state '%s'", state)
 		}
 	}
 
-	return nil
+	backoffCfg := backoff.NewExponentialBackOff()
```
aermakov-zalando: Should we have an upper limit here?

pkg/updatestrategy/aws_asg.go
```diff
@@ -317,25 +320,51 @@ func (n *ASGNodePoolsBackend) Terminate(node *Node, decrementDesired bool) error
...
-		switch aws.StringValue(status.InstanceStatuses[0].InstanceState.Name) {
-		case ec2.InstanceStateNameShuttingDown, ec2.InstanceStateNameStopping, ec2.InstanceStateNameTerminated, ec2.InstanceStateNameStopped:
+		switch state {
+		case ec2.InstanceStateNameShuttingDown, ec2.InstanceStateNameStopping:
+			return errors.New("instance shutting down")
+		case ec2.InstanceStateNameTerminated, ec2.InstanceStateNameStopped:
 			return nil
 		default:
-			return err
+			return fmt.Errorf("unexpected instance state '%s'", state)
```
aermakov-zalando: Should this be `backoff.Permanent`?

pkg/updatestrategy/aws_asg.go
mikkeloscar: 15 min is the default which seems reasonable here.

pkg/updatestrategy/aws_asg.go
aermakov-zalando: If there's a default backoff limit, then we could just ignore. It's not like this is going to happen often (or ever), I guess.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #62, opened 2018-06-20
Mark nodes in orphan node pools

aermakov-zalando:
* `NodePoolManager`: remove `LabelNode`/`TaintNode`
 * `NodePoolManager`: add `MarkNodeForDecommission`, set `lifecycle-status=draining` internally in `drain()`
 * `ScalePool()`: allow resizing to either 0 or size which is gte the current size. The current downscaling logic isn't robust enough (e.g. it doesn't attempt to balance failure domains), so let's only cover the case where a pool disappears completely. Additionally, mark the nodes for decommissioning before downscaling.
 * Log when we actually start terminating a node


mikkeloscar:
:+1:

PR #65, opened 2018-07-18
Pool scale-up: increase empty ASGs last

aermakov-zalando:
Some of the clusters are essentially impossible to upgrade because they have a split pool of spot nodes with no capacity in one of the zones. When CLM tries to scale this cluster up, it increases the currently empty ASG first and then waits forever. This is overall a shitty situation that needs to be fixed properly, but a quick hacky fix would be to prefer scaling up the ASGs that actually have some instances (or requests).

### DIFF
--- a/pkg/updatestrategy/aws_asg.go
+++ b/pkg/updatestrategy/aws_asg.go
@@ -142,10 +142,19 @@ func (n *ASGNodePoolsBackend) Scale(nodePool *api.NodePool, replicas int) error 
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Can you please add an issue for the "real" solution so it's tracked?

mikkeloscar:
:+1:

aermakov-zalando:
👍

aermakov-zalando:
@mikkeloscar sure!

PR #75, opened 2018-08-14
Correctly handle Spot pools in rolling_update

aermakov-zalando:
* Move the special handling for Spot pools to `Update()` to make it more clear.
* Fix the issue where `increaseByUnmatchedNodes` also caused the Spot pool to scale out during updates.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #76, opened 2018-08-14
Use go modules for vendoring

mikkeloscar:
This PR is to play around with [Go Modules](https://github.com/golang/go/wiki/Modules) which we may want to switch to when Go 1.11 lands.


channel/directory_test.go
```diff
@@ -15,7 +15,7 @@ func TestDirectoryChannel(t *testing.T) {
 	d := NewDirectory(location)
 	channels, err := d.Update(logger)
 	require.NoError(t, err)
-	require.NotEmpty(t, channels)
+	require.Empty(t, channels)
```
aermakov-zalando: Why?

pkg/updatestrategy/node_pool_manager.go
```diff
@@ -208,7 +208,7 @@ func (m *KubernetesNodePoolManager) taintNode(node *Node, taintKey, taintValue s
 		return nil
 	}
 
-	backoffCfg := backoff.WithMaxTries(backoff.NewConstantBackOff(1*time.Second), maxConflictRetries)
+	backoffCfg := backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), maxConflictRetries)
```
szuecs: Are these changes triggered by version updates of dependencies?

mikkeloscar: Yes, sorry, I'll do a separate PR with dependency updates to not have everything included here.

mikkeloscar:
:+1:

aermakov-zalando:
👍

aermakov-zalando:
Could you pin `k8s.io/api` to a specific version?

mikkeloscar:
@aermakov-zalando the tags used in `k8s.io/api` are not semver which is required by Go modules. Instead it resolves the tag to the commit and uses that instead. I have now pinned it to the commit which is also used in the version of [client-go that we import](https://github.com/kubernetes/client-go/blob/7d04d0e2a0a1a4d4a1cd6baa432a2301492e4e65/Godeps/Godeps.json#L275).

mikkeloscar:
:+1:

aermakov-zalando:
@mikkeloscar oh well :(

aermakov-zalando:
👍

PR #80, opened 2018-08-16
Persistent draining

aermakov-zalando:
Implement the new node draining algorithm that's persistent, can proceed if CLM is restarted or fails and deals with bad PDBs much better.

The new algorithm works like this:
 1. Mark the node in the usual way, persist the timestamp in an annotation on the node object.
 2. Try evicting all pods on the node in parallel using the eviction API. Continue trying until none of the pods can be evicted because of PDB issues.
 3. Every N seconds (10 for now) update the list of evictable pods and try to evict normally first. If that still doesn't succeed, terminate it forcefully if all the following conditions are met:
      * `-drain-grace-period` has passed since we started draining
      * pod was created at least `-drain-min-pod-lifetime` ago
      * pods in the same PDB were created at least `-drain-min-healthy-sibling-lifetime` ago (if they're healthy) or `-drain-min-unhealthy-sibling-lifetime` (if they're not)
      * last forced termination on the node happened at least `-drain-force-evict-interval` ago
  4. If there was a forced termination, persist the timestamp in an annotation on the node object.

### DIFF
--- a/pkg/updatestrategy/drain.go
+++ b/pkg/updatestrategy/drain.go
@@ -439,6 +439,11 @@ func (m *KubernetesNodePoolManager) isEvictablePod(pod v1.Pod) bool {
 		return false
 	}
 
+	if pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed {
+		logger.Debug("Pod terminated")
+		return false
+	}
+
 	for _, owner := range pod.GetOwnerReferences() {
 		if owner.Kind == "DaemonSet" {
 			logger.Debug("DaemonSet Pod not evictable")
--- a/pkg/updatestrategy/drain.go
+++ b/pkg/updatestrategy/drain.go
@@ -216,7 +216,7 @@ func (m *KubernetesNodePoolManager) forceTerminationAllowed(pod v1.Pod, now, dra
 	if err != nil {
 		return false, err
 	}
-	siblingPods := findSiblingPods(m.logger, pod, allPods, allPdbs)
+	siblingPods := findSiblingPods(m.logger, pod, allPods.Items, allPdbs.Items)
 
 	// check if PDB siblings are old enough
 	siblingsOldEnough := true
@@ -251,12 +251,12 @@ func podReady(pod v1.Pod) bool {
 	return false
 }
 
-func findSiblingPods(logger *log.Entry, pod v1.Pod, allPods *v1.PodList, allPdbs *policy.PodDisruptionBudgetList) []v1.Pod {
+func findSiblingPods(logger *log.Entry, pod v1.Pod, allPods []v1.Pod, allPdbs []policy.PodDisruptionBudget) []v1.Pod {
 	var siblingSelectors []labels.Selector
-	for _, pdb := range allPdbs.Items {
+	for _, pdb := range allPdbs {
 		selector, err := metav1.LabelSelectorAsSelector(pdb.Spec.Selector)
 		if err != nil {
-			logger.Debugf("pdb %s/%s has an invalid selector, skipping", pod.GetNamespace(), pdb.GetName())
+			logger.Debugf("pdb %s/%s has an invalid selector: %v", pod.GetNamespace(), pdb.GetName(), err)
 			continue
 		}
 		if selector.Matches(labels.Set(pod.Labels)) {
@@ -265,7 +265,7 @@ func findSiblingPods(logger *log.Entry, pod v1.Pod, allPods *v1.PodList, allPdbs
 	}
 
 	var result []v1.Pod
-	for _, candidate := range allPods.Items {
+	for _, candidate := range allPods {
 		if candidate.GetName() == pod.GetName() {
 			continue
 		}
--- a/pkg/updatestrategy/drain.go
+++ b/pkg/updatestrategy/drain.go
@@ -175,7 +175,7 @@ func (m *KubernetesNodePoolManager) evictOrForceTerminatePod(ctx context.Context
 		// PDB violation, check if we can force terminate the pod
 		m.pdbViolated(pod)
 
-		forceTerminate, err := m.forceTerminationAllowed(pod, drainStart, lastForcedTermination)
+		forceTerminate, err := m.forceTerminationAllowed(pod, time.Now(), drainStart, lastForcedTermination)
 		if forceTerminate {
 			err = deletePod(m.kube, m.podLogger(pod), pod)
 			if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+
+func (m *KubernetesNodePoolManager) podLogger(pod v1.Pod) *log.Entry {
+	return m.logger.WithFields(log.Fields{
+		"ns":   pod.Namespace,
+		"pod":  pod.Name,
+		"node": pod.Spec.NodeName,
+	})
+}
+
+func (m *KubernetesNodePoolManager) pdbViolated(pod v1.Pod) {
```
szuecs: pdbViolated -> logPdbViolated

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+		// try evicting normally
+		err = evictPod(m.kube, m.logger, pod)
+		if err == nil {
+			return false, nil
+		}
+		if !isPDBViolation(err) {
+			return false, err
+		}
+
+		// PDB violation, check if we can force terminate the pod
```
szuecs: the comment seems wrong to me, because the next part is just logging

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+	return false
+}
+
+func findSiblingPods(logger *log.Entry, pod v1.Pod, allPods []v1.Pod, allPdbs []policy.PodDisruptionBudget) []v1.Pod {
+	var siblingSelectors []labels.Selector
+	for _, pdb := range allPdbs {
+		selector, err := metav1.LabelSelectorAsSelector(pdb.Spec.Selector)
+		if err != nil {
+			logger.Debugf("pdb %s/%s has an invalid selector: %v", pod.GetNamespace(), pdb.GetName(), err)
```
szuecs: Why not use `m.podLogger(pod).Debugf` ?

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+				break
+			}
+		}
+	}
+
+	return result
+}
+
+func isPDBViolation(err error) bool {
+	return apiErrors.IsTooManyRequests(err) || strings.Contains(err.Error(), multiplePDBsErrMsg)
```
szuecs: Do I understand correctly that if apiserver returns 429 we will count this as PDB violation?
Why?

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+}
+
+var deletePod = func(client kubernetes.Interface, logger *log.Entry, pod v1.Pod) error {
+	err := client.CoreV1().Pods(pod.Namespace).Delete(pod.Name, &metav1.DeleteOptions{
+		GracePeriodSeconds: pod.Spec.TerminationGracePeriodSeconds,
+	})
+	if err != nil {
+		logger.Errorf("Failed to delete pod: %v", err)
```
szuecs: Better use `m.podLogger(pod).Errorf`

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+		return err
+	}
+
+	// wait for pod to be terminated and gone from the node.
+	err = waitForPodTermination(client, pod)
+	if err != nil {
+		logger.Warnf("Pod not terminated within grace period: %s", err)
```
szuecs: either error or info. Warn should be only for deprecations.

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+		return err
+	}
+	localLogger.Info("Evicting pod")
+
+	// wait for the pod to be actually evicted and gone from the node.
+	// It has TerminationGracePeriodSeconds time to clean up.
+	start := time.Now().UTC()
+	err = waitForPodTermination(client, pod)
+	if err != nil {
+		localLogger.Warnf("Pod not terminated within grace period: %s", err)
```
szuecs: info or error, not warn, which should be only used for deprecations

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	return result, nil
+}
+
+func (m *KubernetesNodePoolManager) getPodsByNamespace(namespace string) (*v1.PodList, error) {
+	return m.kube.CoreV1().Pods(namespace).List(metav1.ListOptions{})
+}
+
+func (m *KubernetesNodePoolManager) getPDBsbyNamespace(namespace string) (*policy.PodDisruptionBudgetList, error) {
```
szuecs: getPDBsbyNamespace -> getPDBsByNamespace

pkg/updatestrategy/drain_test.go
```diff
@@ -0,0 +1,503 @@
...
+		Spec: v1.PodSpec{
+			NodeName: testNodeName,
+		},
+		Status: v1.PodStatus{
+			Conditions: []v1.PodCondition{{Type: v1.PodReady, Status: status}},
+		},
+	}
+}
+
+func TestForceTerminationGracePeriod(t *testing.T) {
```
szuecs: You can put this + the next 2 tests together and build a so called "table driven test".

Example:
``` 
func TestSkipperCustomRoutes(t *testing.T) {
	for _, ti := range []struct {
		msg            string
                pods             []*v1.Pod
		expected    bool
	}{{
		msg: "test1",
		pods: []*v1.Pod{pod("foo-0", zeroTime, true)},
		expected: true,
        }, {
		msg: "test2",
		pods: []*v1.Pod{pod("foo-0", zeroTime, true)},
		expected: false,

        }, {
		msg: "test3",
                  ...
	}} {
		t.Run(ti.msg, func(t *testing.T) {
			.... result = ..
                        if ti.expected != result {
                            ti.Errorf("...")
                        }  
		})
	}
}
```

pkg/updatestrategy/node_pool_manager.go
```diff
@@ -140,17 +158,75 @@ func (m *KubernetesNodePoolManager) MarkNodeForDecommission(node *Node) error {
...
+		// re-fetch the node since we're going to do an update
+		updatedNode, err := m.kube.CoreV1().Nodes().Get(node.Name, metav1.GetOptions{})
 		if err != nil {
-			return err
+			return backoff.Permanent(err)
 		}
+
+		if patch(updatedNode) {
+			_, err := m.kube.CoreV1().Nodes().Update(updatedNode)
+			if err != nil {
```
szuecs: can be rewritten to the following good practice in Go:

```
if _, err := m.kube.CoreV1().Nodes().Update(updatedNode); err != nil {
...
}
```

provisioner/clusterpy.go
```diff
@@ -615,6 +619,18 @@ func (p *clusterpyProvisioner) prepareProvision(logger *log.Entry, cluster *api.
 	return adapter, updater, poolManager, nil
 }
 
+func handleDurationItem(configItems map[string]string, key string, set func(duration time.Duration)) error {
+	value, ok := configItems[key]
+	if ok {
```
szuecs:
```
if value, ok := configItems[key]; ok {
...
}
```

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+			return err
+		}
+
+		pods, err := m.evictablePods(node.Name)
+		if err != nil {
+			return err
+		}
+
+		if len(pods) == 0 {
+			break
```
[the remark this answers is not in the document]
aermakov-zalando: Good point.

pkg/updatestrategy/drain.go
aermakov-zalando: It's a function, not a method, which makes it a bit easier to test. The logger passed to it already is a `podLogger` which achieves the same behaviour in the end.

pkg/updatestrategy/drain.go
aermakov-zalando: `logger` _is_ a `podLogger`.

pkg/updatestrategy/drain_test.go
aermakov-zalando: I think it's easier to deal with separate test cases as long as there's not a lot of duplicate code. The error messages you get are more clear and you don't have to manually reimplement the logic.

pkg/updatestrategy/drain.go
```diff
@@ -0,0 +1,450 @@
...
+
+func timestampAnnotation(logger *log.Entry, node *Node, annotation string, fallback time.Time) time.Time {
+	ts, ok := node.Annotations[annotation]
+	if !ok {
+		return fallback
+	}
+
+	value, err := time.Parse(time.RFC3339, ts)
+	if err != nil {
+		logger.Warnf("invalid value for %s: %v", annotation, err)
```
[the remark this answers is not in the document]
aermakov-zalando: I don't see anything wrong with using warnings for situations where something is wrong but we can still proceed.

pkg/updatestrategy/drain.go
aermakov-zalando: See above, in my opinion warnings are fine.

pkg/updatestrategy/drain.go
aermakov-zalando: See above, in my opinion warnings are fine.

pkg/updatestrategy/drain.go
aermakov-zalando: This is the check we already had, so nothing changed here. We could also check the error message, but it's easier to just retry.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #82, opened 2018-08-18
Update dependencies

mikkeloscar:
Updates dependencies

* client-go updated to 8.0.0
* go-swagger updated to 0.16.0
* testify updated and one test changed because of the semantics of
`require.NotEmpty` changed in the new version (https://github.com/stretchr/testify/commit/88a414d0725774f245f045f12aee6b94ea66e2f6#diff-8592448e051dac4c311b0cfee177add7R829)

This was done in #76 but it's better to do it separately. We were using quite
old versions of libraries, that's the reason for doing the update at all, but
it also makes introducing Go modules (#76) simpler.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,44 +2,66 @@
 
 
 [[projects]]
+  digest = "1:d1665c44bd5db19aaee18d1b6233c99b0b9a986e8bccb24ef54747547a48027f"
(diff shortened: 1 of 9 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #83, opened 2018-08-20
Downgrade container-linux-config-transpiler

mikkeloscar:
Upgrading the library in #82 had the side effect of changing the
ignition format causing all nodes to be rolled on next update. Revert
the change to prevent this side effect.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #86, opened 2018-09-07
Log why EC2 termination failed

mikkeloscar:
We are swallowing the error in some cases where we shouldn't. At least log the error so we know why it happened.


mikkeloscar:
:+1:

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #87, opened 2018-09-07
Retry instance termination in the exponential backoff

arjunrn:
This PR retries the call to terminate the instance after checking the instance state. This prevents the case when the call is only tried one after which only the state is checked.

### DIFF
--- a/pkg/updatestrategy/aws_asg.go
+++ b/pkg/updatestrategy/aws_asg.go
@@ -3,7 +3,6 @@ package updatestrategy
 import (
 	"errors"
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"math"
 	"sort"
 	"strconv"
@@ -11,6 +10,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/autoscaling/autoscalingiface"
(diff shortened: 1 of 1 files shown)


pkg/updatestrategy/aws_asg.go
```diff
@@ -339,23 +340,32 @@ func (n *ASGNodePoolsBackend) Terminate(node *Node, decrementDesired bool) error
...
+					default:
+						return backoff.Permanent(err)
+					}
+				} else {
+					// at this point the call to the API failed. try later
+					return err
+				}
+			} else {
+				return errors.New("signalled instance to shutdown")
+			}
```
mikkeloscar: No need for the two `else` since you return in the if in both cases :)

pkg/updatestrategy/aws_asg.go
```diff
@@ -364,9 +374,10 @@ func (n *ASGNodePoolsBackend) Terminate(node *Node, decrementDesired bool) error
 			return fmt.Errorf("unexpected instance state '%s'", state)
 		}
 	}
-
+	// wait for the instance to be terminated/stopped
 	backoffCfg := backoff.NewExponentialBackOff()
-	return backoff.Retry(instanceState, backoffCfg)
+	backoffCfg.MaxElapsedTime = time.Duration(15) * time.Minute
```
mikkeloscar: nit: Could you move this timeout to a `const` in the top?

pkg/updatestrategy/aws_asg.go
arjunrn: The reason I wrote it like that is so that there are different errors returns for different situations. I thought it would be good debugging information. But turns out the `backoff` library doesn't ever log anything so it doesn't matter. Will change.

pkg/updatestrategy/aws_asg.go
```diff
@@ -17,20 +19,20 @@ import (
...
+	clusterIDTagPrefix               = "kubernetes.io/cluster/"
+	resourceLifecycleOwned           = "owned"
+	kubeAutoScalerEnabledTagKey      = "k8s.io/cluster-autoscaler/enabled"
+	nodePoolTag                      = "NodePool"
+	userDataAttribute                = "userData"
+	instanceTypeAttribute            = "instanceType"
+	instanceIdFilter                 = "instance-id"
+	instanceHealthStatusHealthy      = "Healthy"
+	ec2AutoscalingGroupTagKey        = "aws:autoscaling:groupName"
+	instanceTerminationRetryDuration = 15
```
mikkeloscar: I meant that the full const i.e. `instanceTerminationRetryDuration = 15 * time.Minute` should be defined here. :)

arjunrn:
:+1:

mikkeloscar:
:+1:

PR #90, opened 2018-09-18
findPDBSiblings: ignore terminated pods & job pods

aermakov-zalando:
Our users often have a PDB which matches by application name and then create a deployment & a bunch of cronjobs with the same application. This leads to an invalid PDB, but since we don't know anything about that, we'll continue waiting forever because there will always be fresh PDB siblings.

Let's just ignore Job pods and terminated pods for now in the PDB sibling search function. If this doesn't work either, we could also introduce a hard timeout on PDB siblings, but let's see first.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #91, opened 2018-09-21
Fail on invalid manifests

aermakov-zalando:
* Render all manifests before doing anything to the cluster, returning an error if one of them is invalid
 * Return an error if we can't apply a manifest instead of continuing


arjunrn:
:+1:

aermakov-zalando:
👍

PR #92, opened 2018-09-21
Fix manifest path in renderManifests()

aermakov-zalando:
It was a relative path, not an absolute one.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #94, opened 2018-10-09
Add generateCertificate template function

aermakov-zalando:
Add a `generateCertificate` template function that generates and signs a certificate using a provided CA.


aermakov-zalando:
Closed because this will cause the nodes to be rotated every time, so it's better if we solve it in some other way.

PR #96, opened 2018-11-01
Better instance storage support

aermakov-zalando:
* Instance info: add `NVMEInstanceStorage`
* Node pools: expose `instance_info` in `.Values`
* Add `mountUnitName` template function


aermakov-zalando:
👍

PR #97, opened 2018-11-02
Use a $100/hr spot price cap if no pricing info is available

aermakov-zalando:
This allows us to proceed even if we don't have pricing info and *should* be reasonably harmless. We'll move to launch templates soon anyway, so this is just a temporary hack.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #101, opened 2018-11-28
Add a separate concurrency limit for forks

### LINKED ISSUE #98: Memory usage causes errors and panics
CLM forks external processes when it updates the clusters (`senza` to prepare the template, `kubectl` to apply the manifests). The processes usually consume significant amount of memory, which means that if we use a large number of workers most of the calls will fail (or, even worse, stuff inside CLM will begin `panic`-ing because it can't allocate memory). However, we also don't want to reduce the number of workers because then we'll have to wait for ages to roll out the updates. Since most of the time is actually spent waiting in `drain()`, we should just set a separate limit of the number of external processes instead.

aermakov-zalando:
Add a new argument, `--concurrent-external-processes`, which limit the number of external processes (like `senza` or `kubectl`) we run at the same time.
Fix #98.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -811,9 +811,12 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:39ebcc2b11457b703ae9ee2e8cca0f68df21969c6102cb3b705f76cca0ea0239"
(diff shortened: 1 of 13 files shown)


mikkeloscar:
:+1:

szuecs:
:+1:

PR #102, opened 2018-11-30
Allow relative paths to to kubernetes on aws when running from within its path

dionb:
Signed-off-by: Dion Bramley <dion.bramley@zalando.de>


dionb:
:+1:

mikkeloscar:
A test case for this would be good to have :)

Especially since `baseDir` kind of looses it's meaning when we do a `Contains` check.

mikkeloscar:
Closing for now as this has not been worked on.

PR #105, opened 2018-12-05
Close the context once we're done

aermakov-zalando:
The context was leaking since it was only closed if the update was aborted.


szuecs:
:+1:

aermakov-zalando:
👍

PR #106, opened 2018-12-06
CLC update strategy

aermakov-zalando:
MVP for Cluster Lifecycle Controller integration. If update strategy is set to `clc`, only update the launch configuration & launch templates, mark the outdated nodes with `decommission-pending` and wait for CLC to remove them.

Note that CLM also does the draining when decommissioning node pools or scaling down; this PR only touches the node update code. Once we're comfortable with using CLC, I'd rather drop the old code completely and refactor the whole `node_pool_manager`/`update_strategy` mess.

### DIFF
--- a/pkg/updatestrategy/clc_update.go
+++ b/pkg/updatestrategy/clc_update.go
@@ -1,0 +1,107 @@
+package updatestrategy
+
+import (
+	"context"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
+)
+
+const (
+	pollingInterval = 10 * time.Second
+)
+
+type CLCUpdateStrategy struct {
+	nodePoolManager NodePoolManager
+	logger          *log.Entry
+}
+
+// NewCLCUpdateStrategy initializes a new CLCUpdateStrategy.
+func NewCLCUpdateStrategy(logger *log.Entry, nodePoolManager NodePoolManager) *CLCUpdateStrategy {
+	return &CLCUpdateStrategy{
+		nodePoolManager: nodePoolManager,
+		logger:          logger.WithField("strategy", "clc"),
+	}
+}
+
+func (c *CLCUpdateStrategy) Update(ctx context.Context, nodePoolDesc *api.NodePool) error {
+	c.logger.Infof("Initializing update of node pool '%s'", nodePoolDesc.Name)
+
+	err := c.doUpdate(ctx, nodePoolDesc)
+	if err != nil {
+		if ctx.Err() != nil {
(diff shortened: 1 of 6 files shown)


provisioner/clusterpy.go
```diff
@@ -621,47 +622,50 @@ func (p *clusterpyProvisioner) prepareProvision(logger *log.Entry, cluster *api.
...
+		{"drain_poll_interval", func(v time.Duration) { drainConfig.PollInterval = v }},
+	} {
+		if value, ok := cluster.ConfigItems[setting.key]; ok {
+			parsed, err := time.ParseDuration(value)
+			if err != nil {
+				return nil, nil, nil, fmt.Errorf("invalid value for %s: %v", setting.key, err)
+			}
+			setting.fn(parsed)
+		}
+	}
```
mikkeloscar: I'm not sure this change makes it easier to read the code?

pkg/updatestrategy/clc_update.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func (c *CLCUpdateStrategy) Update(ctx context.Context, nodePoolDesc *api.NodePool) error {
+	c.logger.Infof("Initializing update of node pool '%s'", nodePoolDesc.Name)
+
+	err := c.doUpdate(ctx, nodePoolDesc)
+	if err != nil {
+		if ctx.Err() != nil {
+			err := c.rollbackUpdate(nodePoolDesc)
```
mikkeloscar: Is the intention that the CLC will handle this in the future? I understand why it's done this way, but it seems like the CLC should do this by itself at some point or is that not the intention?

provisioner/clusterpy.go
aermakov-zalando: The change makes it actually handle the errors, previously it ignored those completely. :(

pkg/updatestrategy/clc_update.go
aermakov-zalando: I don't really have any plans for this, we could move cluster-update blocking to the clusters as well I guess?

pkg/updatestrategy/clc_update_test.go
```diff
@@ -0,0 +1,230 @@
+package updatestrategy
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
```
szuecs: please clean empty lines

aermakov-zalando: Fixed.

mikkeloscar:
:+1:

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #114, opened 2019-01-04
Drop support for LaunchConfigurations

aermakov-zalando:
We don't use them anymore, and it'll just complicate the code responsible for MixedInstancesPolicy support.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #115, opened 2019-01-07
Support for multiple instance types (MixedInstancesPolicy)

aermakov-zalando:
Add support for ASGs with multiple instance types (using MixedInstancesPolicy):
 * node pools: add `instance_types`, keep `instance_type` set to the first type for BC
 * support ASGs that use `MixedInstancesPolicy`:
   - fetch launch template name & version correctly
   - determine whether the ASG uses spot or not (on-demand+spot not supported at the moment)
   - determine whether the instances should be updated (e.g. because one of the types is no longer valid)
 * template values:
   - `supports_t2_unlimited`: set to true iff all instance types are T2
   - `instance_info`: if a pool uses more than one type, return a synthetic value with instance type set to `<multiple>`, CPU/Memory set to the minimum across all types. Storage information must be the same across all types, an error will be thrown if it's not the case.
 * add a bunch of tests

### DIFF
--- a/registry/file.go
+++ b/registry/file.go
@@ -4,10 +4,9 @@ import (
 	"fmt"
 	"io/ioutil"
 
-	yaml "gopkg.in/yaml.v2"
-
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
+	"gopkg.in/yaml.v2"
 )
 
 type fileRegistry struct {
@@ -41,6 +40,9 @@ func (r *fileRegistry) ListClusters(filter Filter) ([]*api.Cluster, error) {
 
 	for _, cluster := range fileClusters.Clusters {
 		for _, nodePool := range cluster.NodePools {
+			if len(nodePool.InstanceTypes) == 0 {
+				return nil, fmt.Errorf("no instance types for cluster %s, pool %s", cluster.ID, nodePool.Name)
+			}
 			nodePool.InstanceType = nodePool.InstanceTypes[0]
 		}
 	}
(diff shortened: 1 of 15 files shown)


registry/file.go
```diff
@@ -39,6 +39,12 @@ func (r *fileRegistry) ListClusters(filter Filter) ([]*api.Cluster, error) {
 		return nil, err
 	}
 
+	for _, cluster := range fileClusters.Clusters {
+		for _, nodePool := range cluster.NodePools {
+			nodePool.InstanceType = nodePool.InstanceTypes[0]
```
mikkeloscar: Do we want to check that `len(InstanceTypes) > 0`?

registry/http.go
```diff
@@ -178,7 +178,8 @@ func convertFromClusterModel(cluster *models.Cluster) *api.Cluster {
 func convertFromNodePoolModel(nodePool *models.NodePool) *api.NodePool {
 	return &api.NodePool{
 		DiscountStrategy: *nodePool.DiscountStrategy,
-		InstanceType:     *nodePool.InstanceType,
+		InstanceTypes:    nodePool.InstanceTypes,
+		InstanceType:     nodePool.InstanceTypes[0],
```
mikkeloscar: Same here! Check `len(InstanceTypes) > 0`

aermakov-zalando: Fixed.

registry/file.go
aermakov-zalando: Fixed.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #116, opened 2019-01-16
Fix pool draining

aermakov-zalando:
When draining, remove nodes using the `UpdateStrategy` to avoid concurrent removal by both CLC and CLM. The code is a complete mess WRT to abstraction layers, but I'd rather refactor it after we switch to CLC in all clusters and drop the whole concept of update strategies.


mikkeloscar:
:+1:

PR #117, opened 2019-01-22
Buffer pods: use pod count when calculating the size

aermakov-zalando:
Update `autoscalingBufferSettings` to pay attention to `autoscaling_buffer_pods` as well. If it's 0, return a dummy `100m`/`100Mi` size immediately without checking other settings, otherwise divide the buffer size by the expected number of replicas. This makes the evictions a bit more fine-grained (so not the whole node is vacated at once) and side-steps the issue of having to update the settings every time we change the daemonsets.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -20,6 +20,7 @@ import (
 )
 
 const (
+	autoscalingBufferPodCountConfigItem       = "autoscaling_buffer_pods"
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
On the other hand, this will also mean that the buffer pods will spread around instances and there won't be enough space anywhere to fit a large pod from a suddenly terminated node. Closing this for now.

PR #124, opened 2019-02-11
Update AWS instance data

aermakov-zalando:
* `instances.json`: remove pricing & sort by instance type so we have nice diffs
* `instances.json`: update to include t3 instances. Manually updated to [this version](https://github.com/powdahound/ec2instances.info/blob/acd973261819475f3f0765baaedb13c65988d6ce/www/instances.json) because data about instance storage and EBS-as-NVME is completely broken in the latest versions


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #126, opened 2019-02-12
CLC: disable autoscaling when decommissioning pools

aermakov-zalando:
CLC strategy didn't disable autoscaling for pools that were decommissioned, so it never finished.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #130, opened 2019-03-04
CLC: when updating a pool, abort decommissioning in other pools

aermakov-zalando:
When a node pool is being updated, set the update priority annotation to the current time and forcibly abort decommissioning processes in other node pools. This allows us to have behaviour that's close to what we had without CLC, where if we abort an update in progress master nodes will be updated again, even if some of the worker nodes were still being decommissioned.

The code is awful because of mismatch between multiple abstraction layers, but I'd rather just drop most of it completely and rewrite it into something sane once we drop non-CLC clusters completely.


aermakov-zalando:
This is not needed, we can do it in CLC.

PR #131, opened 2019-03-24
Check Etcd CF stack before running senza

mikkeloscar:
This is a simple hack to avoid invoking senza on every run of CLM when
the resulting generated CF stack won't be used anyway.

Should make CLM more robust to AWS ratelimits and avoids senza as
dependency if there's already an etcd cluster in the target AWS account.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #134, opened 2019-03-25
Allow templating in deletions.yaml

aermakov-zalando:
We have some resources that shouldn't be present depending on a config item value, and currently we can't remove them reliably.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #135, opened 2019-03-29
Fixed the check for existing etcd stack if it doesn't exist

arjunrn:
Currently if the stack doesn't exist the CLM doesn't proceed at all because this check fails.

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -410,11 +410,8 @@ func (a *awsAdapter) CreateOrUpdateEtcdStack(parentCtx context.Context, stackNam
 	}
 
 	resp, err := a.cloudformationClient.DescribeStacks(describeParams)
-	if err != nil {
-		return err
-	}
-
-	if len(resp.Stacks) == 1 {
+	// Ignore the error because the error indicates that the stack is missing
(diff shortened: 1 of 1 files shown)


provisioner/aws.go
```diff
@@ -410,11 +410,7 @@ func (a *awsAdapter) CreateOrUpdateEtcdStack(parentCtx context.Context, stackNam
 	}
 
 	resp, err := a.cloudformationClient.DescribeStacks(describeParams)
-	if err != nil {
-		return err
-	}
-
-	if len(resp.Stacks) == 1 {
+	if err ==nil && len(resp.Stacks) == 1 {
```
aermakov-zalando: Please fix the formatting!

aermakov-zalando: I'd also add some comment about why we're ignoring the error (not a big deal and CF API doesn't report non-existent stacks in an obvious way).

njuettner: I raised the same question in person. Please comment it why ignoring an error 🙂.

muaazsaleem:
👍

aermakov-zalando:
👍

PR #138, opened 2019-04-15
Simplify Decommission API

mikkeloscar:
This simplifies the Provisioner Decommission API to not require the channelConfig which is not used anyway. This change is done to simplify pulling in CLM as a library which I'm using for our internal e2e test cleaner job.

Additionally it checks in the `pkg/cluster-registry/models` generated package as this is imported in `provisioner/clusterpy.go` and therefore needed if you want to import that as a library.


mikkeloscar:
:+1:

njuettner:
👍

PR #140, opened 2019-04-18
Unify template data, add proper AZ support

aermakov-zalando:
Add support for `availability_zones` property. It can be specified at either the cluster level, in which case it affects everything, or on the node pool level (then it's merged with the cluster values). Subnets are automatically restricted to only the available AZs, and `values` now gains a new attribute, `availability_zones`.

Template data is now unified across all templates, which makes it possible to use `values` in Kubernetes manifests as well. Some of the data (e.g. `NodePool`) will obviously not be available in all of the templates, but at least the structure of the data will be the same. Ideally, we should migrate all Kubernetes manifests to use `.Cluster.ConfigItems` instead of `.ConfigItems` and then drop the compatibility properties in `templateData`.

### DIFF
--- a/provisioner/aws_az.go
+++ b/provisioner/aws_az.go
@@ -1,0 +1,50 @@
+package provisioner
+
+import (
+	"sort"
+)
+
+// AZInfo tracks information about available AZs based on explicit restrictions or available subnets
+type AZInfo struct {
+	subnets map[string]string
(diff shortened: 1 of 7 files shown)


njuettner:
👍

patandrick:
:+1:

PR #142, opened 2019-05-22
Support for cloud-init and CLC at the same time

linki:
This adds back support for plain-old `cloud-init` in parallel with `clc/ignition`.

Upon defining the NodePool's UserData CLM looks for the current `userdata.clc.yaml` and processes/uploads it as before. If the file doesn't exist it checks for a file `userdata.yaml`, processes it and sets its content as UserData.

Note, that it doesn't upload to S3 at this point as it's not as easy as with `ignition` afaik. With a lot of stuff moving into the custom Ubuntu AMI it's likely we manage to stay under the UserData limit this time, WDYT?

Code is not the prettiest but as I understood this conditional is temporary again? If not I'm happy to make it a little nicer.

/cc @njuettner @mikkeloscar  @aermakov-zalando

### DIFF
--- a/provisioner/node_pools.go
+++ b/provisioner/node_pools.go
@@ -255,8 +255,8 @@ func (p *AWSNodePoolProvisioner) prepareUserData(templateCtx *templateContext, n
 
 // prepareCloudInit prepares the user data by rendering the golang template.
 // A EC2 UserData ready base64 string will be returned.
-func (p *AWSNodePoolProvisioner) prepareCloudInit(templateCtx *templateContext, clcPath string) (string, error) {
-	rendered, err := renderTemplate(templateCtx, clcPath)
+func (p *AWSNodePoolProvisioner) prepareCloudInit(templateCtx *templateContext, cloudInitPath string) (string, error) {
+	rendered, err := renderTemplate(templateCtx, cloudInitPath)
 	if err != nil {
 		return "", err
 	}
--- a/provisioner/node_pools.go
+++ b/provisioner/node_pools.go
@@ -23,7 +23,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
LGTM, thanks!

Long term I'd still upload it to S3 (and encrypt with KMS), but for now it should be fine.

provisioner/node_pools.go
```diff
@@ -235,10 +234,40 @@ func (p *AWSNodePoolProvisioner) Reconcile(ctx context.Context, updater updatest
...
+	if _, err := os.Stat(cloudInitPath); err == nil {
+		return p.prepareCloudInit(templateCtx, cloudInitPath)
+	}
+
+	return "", fmt.Errorf("no userdata file at '%s' nor '%s' found", clcPath, cloudInitPath)
+}
+
+// prepareCloudInit prepares the user data by rendering the golang template.
+// A EC2 UserData ready base64 string will be returned.
+func (p *AWSNodePoolProvisioner) prepareCloudInit(templateCtx *templateContext, clcPath string) (string, error) {
```
njuettner: clcPath needs to be renamed to cloudInitPath

njuettner:
👍

mikkeloscar:
> With a lot of stuff moving into the custom Ubuntu AMI it's likely we manage to stay under the UserData limit this time, WDYT?

I'm afraid we can't fit master config into userdata. The yamls alone are more than 16K :(  Ofc. we could put some of this into the AMI but would limit/complicate our ability to do custom feature flags etc.

I have nothing against merging this, we may just need to rethink this part to handle master config once we get a little further with the AMI.

njuettner:
👍

linki:
👍

PR #145, opened 2019-05-28
Cloud-init gzip and generated files upload to S3

arjunrn:
Added functionality so that the cloud-init userdata file is gzipped and PKI is uploaded to S3

### DIFF
--- a/provisioner/node_pools.go
+++ b/provisioner/node_pools.go
@@ -1,7 +1,9 @@
 package provisioner
 
 import (
+	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"context"
 	"crypto/sha512"
 	"encoding/base64"
@@ -20,16 +22,19 @@ import (
 	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
 	awsExt "github.com/zalando-incubator/cluster-lifecycle-manager/pkg/aws"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/updatestrategy"
+	"gopkg.in/yaml.v2"
 )
 
 const (
 	clcFileName           = "userdata.clc.yaml"
 	cloudInitFileName     = "userdata.yaml"
 	stackFileName         = "stack.yaml"
+	filesTemplateName     = "files.yaml"
 	nodePoolTagKeyLegacy  = "NodePool"
 	nodePoolTagKey        = "kubernetes.io/node-pool"
 	nodePoolRoleTagKey    = "kubernetes.io/role/node-pool"
 	nodePoolProfileTagKey = "kubernetes.io/node-pool/profile"
+	defaultSSLPermissions = 0400
 )
 
 // NodePoolProvisioner is able to provision node pools for a cluster.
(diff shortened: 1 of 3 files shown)


provisioner/node_pools.go
```diff
@@ -247,21 +252,36 @@ func (p *AWSNodePoolProvisioner) prepareUserData(templateCtx *templateContext, n
...
 	}
-
-	return base64.StdEncoding.EncodeToString([]byte(rendered)), nil
+	var gzipBuffer bytes.Buffer
+	writer := gzip.NewWriter(&gzipBuffer)
+	_, err = writer.Write([]byte(rendered))
+	if err != nil {
+		return "", err
+	}
+	err = writer.Close()
```
szuecs: too late, please use defer after line 275

provisioner/node_pools.go
```diff
@@ -317,6 +337,46 @@ func (p *AWSNodePoolProvisioner) uploadUserDataToS3(userData []byte, bucketName
...
+			return "", err
+		}
+		tarWriter.WriteHeader(&tar.Header{
+			Name: key,
+			Size: int64(len(decoded)),
+			Mode: defaultSSLPermissions,
+		})
+		tarWriter.Write(decoded)
+	}
+	err = tarWriter.Close()
```
szuecs: too late, you would not close on error of the for loop for example please use defer

provisioner/node_pools.go
```diff
@@ -317,6 +337,46 @@ func (p *AWSNodePoolProvisioner) uploadUserDataToS3(userData []byte, bucketName
...
+	if err != nil {
+		return "", nil
+	}
+	err = gzw.Close()
```
szuecs: too late, you would not close on error of the for loop for example please use defer

provisioner/node_pools.go
arjunrn: In this case I would  argue it is not because we don't need to close the writer because there is no risk of leaving dangling open file pointer. OTOH if we don't check the the return value of `close()` we might end up with a what appears to be a successful write but has in fact failed.

provisioner/node_pools.go
arjunrn: same argument here

provisioner/node_pools.go
arjunrn: same here

provisioner/node_pools.go
```diff
@@ -247,21 +252,36 @@ func (p *AWSNodePoolProvisioner) prepareUserData(templateCtx *templateContext, n
...
+		return p.prepareCloudInit(nodePoolProfilesPath, templateCtx, cloudInitPath)
 	}
 
 	return "", fmt.Errorf("no userdata file at '%s' nor '%s' found", clcPath, cloudInitPath)
 }
 
 // prepareCloudInit prepares the user data by rendering the golang template.
 // A EC2 UserData ready base64 string will be returned.
-func (p *AWSNodePoolProvisioner) prepareCloudInit(templateCtx *templateContext, cloudInitPath string) (string, error) {
+func (p *AWSNodePoolProvisioner) prepareCloudInit(nodePoolProfilesPath string, templateCtx *templateContext, cloudInitPath string) (string, error) {
```
linki: could you update the docs as well, i.e. s3 upload, tar, gzip

provisioner/node_pools.go
```diff
@@ -317,6 +338,48 @@ func (p *AWSNodePoolProvisioner) uploadUserDataToS3(userData []byte, bucketName
 	return fmt.Sprintf("s3://%s/%s", bucketName, objectName), nil
 }
 
+func (p *AWSNodePoolProvisioner) renderUploadPKI(templateCtx *templateContext, nodePoolProfilePath string) (string, error) {
```
linki: A comment describing the function would be very helpful here.

provisioner/node_pools.go
arjunrn: Looks like `close()` can be called multiple times. So will add it in both places.

provisioner/node_pools.go
```diff
@@ -20,16 +22,19 @@ import (
...
 	awsExt "github.com/zalando-incubator/cluster-lifecycle-manager/pkg/aws"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/updatestrategy"
+	"gopkg.in/yaml.v2"
 )
 
 const (
 	clcFileName           = "userdata.clc.yaml"
 	cloudInitFileName     = "userdata.yaml"
 	stackFileName         = "stack.yaml"
+	pkiFileName           = "pki.yaml"
```
aermakov-zalando: Could you rename it to something more generic? We could use it for other things, not necessarily PKI.

linki:
👍

arjunrn:
:+1:

provisioner/node_pools.go
```diff
@@ -247,21 +252,38 @@ func (p *AWSNodePoolProvisioner) prepareUserData(templateCtx *templateContext, n
...
+// It also uploads the PKI needed for the nodes in the pool
 // A EC2 UserData ready base64 string will be returned.
-func (p *AWSNodePoolProvisioner) prepareCloudInit(templateCtx *templateContext, cloudInitPath string) (string, error) {
+func (p *AWSNodePoolProvisioner) prepareCloudInit(nodePoolProfilesPath string, templateCtx *templateContext, cloudInitPath string) (string, error) {
+	s3Path, err := p.renderUploadPKI(templateCtx, nodePoolProfilesPath)
+	if err != nil {
+		return "", err
+	}
+	templateCtx.s3PKIPath = s3Path
+	p.logger.Debugf("Uploaded PKI to %s", s3Path)
```
aermakov-zalando: This still says "PKI".

aermakov-zalando:
The code still mentions "PKI" in a lot of places.

arjunrn:
:+1:

linki:
👍

PR #146, opened 2019-06-11
Changed file uploader to also encrypt

arjunrn:
Changed the `files.yaml` schema to specify encrypted files and permissions. 

Each of the files are encrypted separately because of the limit set by KMS.

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -19,6 +19,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/aws/aws-sdk-go/service/kms/kmsiface"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/aws/aws-sdk-go/service/sts"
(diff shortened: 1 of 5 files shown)


provisioner/aws.go
```diff
@@ -107,6 +108,10 @@ type s3UploaderAPI interface {
 	Upload(input *s3manager.UploadInput, options ...func(*s3manager.Uploader)) (*s3manager.UploadOutput, error)
 }
 
+type kmsAPI interface {
+	Encrypt(input *kms.EncryptInput) (*kms.EncryptOutput, error)
+}
```
mikkeloscar: Please use the `kmsiface`: https://github.com/aws/aws-sdk-go/tree/master/service/kms/kmsiface We have not done it in other places in this project, but I think it makes more sense than defining a custom interface each time.

Here's an example of stsiface being used in a similar way: https://github.com/mikkeloscar/kube-aws-iam-controller/blob/2c39d429695d6ae624d794fa19e7755c2b9a0112/credentials_getter_test.go#L16

arjunrn: What magic is this :)

aermakov-zalando:
👍

arjunrn:
:+1:

provisioner/remote_files.go
```diff
@@ -0,0 +1,60 @@
...
+	defer gzw.Close()
+	tarWriter := tar.NewWriter(gzw)
+	defer tarWriter.Close()
+	for _, remoteFile := range data.Files {
+		var compressionInput []byte
+		var finalPath string
+
+		decoded, err := base64Decode(remoteFile.Data)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode data: %v for file: %s", remoteFile.Data, remoteFile.Path)
```
mikkeloscar: Maybe we should not log the data in case it's supposed to be encrypted :)

arjunrn: Fix in [PR](https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/147)

PR #147, opened 2019-06-12
Changed log line to not log potentially sensitive data

arjunrn:
Do not log data in case it is supposed to be encrypted.


arjunrn:
:+1:

mikkeloscar:
:+1:

PR #148, opened 2019-06-17
Fixed rolling updates for ubuntu nodes

arjunrn:
Changed the `uploadUserDataToS3` so that it accepts the userdata hash separately because for Ubuntu nodes the userdata is encrypted with KMS which means the ciphertext is different on every run for the same plaintext.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #150, opened 2019-07-10
Be more robust when deleting stacks

mikkeloscar:
When deleting a cluster don't try to delete stacks which are already being deleted. ~disable termination protection of stacks where it's already disabled (and potentially already being deleted).~


szuecs:
:+1:

szuecs:
:+1:

mikkeloscar:
:+1:

PR #152, opened 2019-07-25
Tag main stack with clusterID tag

mikkeloscar:
Tags the main cluster stack with clusterID tag such that it's easier to map the stack to the owning cluster. This will be useful for cleaning up e2e test resources.

It also adds a tag `cluster-lifecycle-controller.zalando.org/main-stack=true` to ensure that we don't find the main stack when list other cluster stacks during cluster deletions (The deletion has to be done in a certain order because of cross-stack references).


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #153, opened 2019-07-26
Propagate cluster config items to node pool

linki:
Forward config items from the cluster to the node pool if it's not defined on the node pool.

Allows to use `{{ .NodePool.ConfigItems.pod_max_pids }}` in the templates without any guarding ifs. Logically, it will lookup the value first in the node pool's config items, then in the cluster's config items, then it uses the default. Allows to have cluster-wide defaults for node-pool specific values such as PID limit or node pool labels.

An alternative would be to - instead - overwrite `Cluster.ConfigItems` with `NodePool.ConfigItems` if duplicate keys exist. In this case we could always use the same syntax for reading config items, i.e. ``{{ .Cluster.ConfigItems.pod_max_pids }}``. This way any config item can be overwritten by any node pool and the templates don't need to know.

Let me know what you prefer.

/cc @aermakov-zalando @mikkeloscar

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -147,6 +147,22 @@ func (p *clusterpyProvisioner) decryptConfigItems(cluster *api.Cluster) error {
 	return nil
 }
 
+// propagateConfigItemsToNodePools propagates cluster-wide config items
+// to each node pool unless the node pool defines its own value.
+func (p *clusterpyProvisioner) propagateConfigItemsToNodePools(cluster *api.Cluster) {
+	for _, nodePool := range cluster.NodePools {
+		// If the node pool doesn't define any config items, we need to initialize it here.
+		if nodePool.ConfigItems == nil {
+			nodePool.ConfigItems = map[string]string{}
+		}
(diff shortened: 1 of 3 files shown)


linki:
Example how templates can be simplified: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2360

aermakov-zalando:
Thanks! I'm not sure if we should do it the second way — having the config item read from `NodePool` gives you a hint that it makes sense to customise it per-node pool, but that's the only downside I could think of. If we decide to keep it this way, could you move this a bit further up? Some future parts of the code could look at node pool config items before `provisionNodePool` is called, and I'd like them to see consistent values. For example, it could be performed near the default initialisation, which is done almost at the start of cluster processing.

mikkeloscar:
I agree that the first option makes most sense because you are "overwriting" in one direction: `defaults` -> `cluster.config_items` -> `node_pool.config_items`.

linki:
@aermakov-zalando Moved it up.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #155, opened 2019-07-31
Added template function to generate AMI id

arjunrn:
Added a new `amiID` function to the template context which can get the AMI ID when given the `name` and `owner-id`.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -20,6 +20,8 @@ import (
 	"strings"
 	"text/template"
 
+	awsUtil "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/aws"
 	k8sresource "k8s.io/apimachinery/pkg/api/resource"
@@ -35,6 +37,8 @@ const (
 	autoscalingBufferPoolsConfigItem          = "autoscaling_buffer_pools"
 	highestPossiblePort                       = 65535
 	lowestPossiblePort                        = 0
+	describeImageFilterNameName               = "name"
+	describeImageFilterNameOwner              = "owner-id"
 )
 
 type templateContext struct {
@@ -46,6 +50,7 @@ type templateContext struct {
 	userData              string
 	computingManifestHash bool
 	s3GeneratedFilesPath  string
+	awsAdapter            *awsAdapter
 }
 
 type templateData struct {
@@ -84,7 +89,7 @@ type podResources struct {
 	Memory string
 }
 
-func newTemplateContext(baseDir string, cluster *api.Cluster, nodePool *api.NodePool, values map[string]interface{}, userData string) *templateContext {
+func newTemplateContext(baseDir string, cluster *api.Cluster, nodePool *api.NodePool, values map[string]interface{}, userData string, adapter *awsAdapter) *templateContext {
 	return &templateContext{
 		baseDir:      baseDir,
 		manifestData: make(map[string]string),
@@ -92,6 +97,7 @@ func newTemplateContext(baseDir string, cluster *api.Cluster, nodePool *api.Node
 		nodePool:     nodePool,
 		values:       values,
 		userData:     userData,
+		awsAdapter:   adapter,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


provisioner/template.go
```diff
@@ -270,6 +275,11 @@ func renderTemplate(context *templateContext, filePath string) (string, error) {
 		"publicKey":                 publicKey,
 		"stupsNATSubnets":           stupsNATSubnets,
 	}
+	if len(extraTemplateFuncs) > 0 {
```
linki: I think we can drop the if here.

provisioner/node_pools.go
```diff
@@ -64,6 +69,38 @@ type AWSNodePoolProvisioner struct {
...
+		ownerAccount := params[1]
+		input := ec2.DescribeImagesInput{Filters: []*ec2.Filter{
+			{Name: &filterNameName, Values: []*string{&name}},
+			{Name: &filterOwnerName, Values: []*string{&ownerAccount}},
+		}}
+		output, err := a.ec2Client.DescribeImages(&input)
+		if err != nil {
+			return "", fmt.Errorf("failed to describe image with name %s and owner %s: %v", name, ownerAccount, err)
+		}
+		if len(output.Images) != 1 {
```
linki: When search for AMIs by name and owner, isn't it expected that we get multiple back from AWS (and then select the most recent one?)

arjunrn: Yes in the normal case but I am assuming here that the image names are immutable because of the pipelines.

linki: So it's something like `kubernetes-image-ami-master-42`?

aermakov-zalando:
Thanks! Is there any reason for doing it only in one case, or was it just simpler to write? I'd prefer all templates to work the same, i.e. add `amiID` function to the default list and just pass the AWS client to `templateContext`. How hard would it be?

provisioner/node_pools.go
arjunrn: Yes

provisioner/template.go
```diff
@@ -552,3 +558,24 @@ func stupsNATSubnets(vpcCidr string) ([]string, error) {
...
+func amiID(adapter *awsAdapter, imageName, imageOwner string) (string, error) {
+	if adapter == nil || adapter.ec2Client == nil {
+		return "", fmt.Errorf("the ec2 client is not available")
+	}
+
+	filterNameName := describeImageFilterNameName
+	filterOwnerName := describeImageFilterNameOwner
+	input := ec2.DescribeImagesInput{Filters: []*ec2.Filter{
+		{Name: &filterNameName, Values: []*string{&imageName}},
+		{Name: &filterOwnerName, Values: []*string{&imageOwner}},
```
linki: You could also use `aws.String()` for that. You might be able to drop the temporary variables as well. There's also [StringSlice](https://godoc.org/github.com/aws/aws-sdk-go/aws#StringSlice) for the values part.

provisioner/template_test.go
```diff
@@ -507,3 +514,72 @@ func TestStupsNATSubnetsErrors(t *testing.T) {
...
+		imageID   string
+		output    []*ec2.Image
+		expectErr bool
+	}{
+		{
+			name:      "basic",
+			imageName: "kubernetes-image-ami",
+			ownerID:   "8085",
+			imageID:   "ami-0001dsf",
+			output:    []*ec2.Image{{ImageId: stringPtr("ami-0001dsf")}},
```
linki: Let's use `aws.String()` like [here](https://github.com/kubernetes-incubator/external-dns/blob/master/provider/aws.go#L485).

aermakov-zalando:
Do you think it makes sense to cache the resolved AMIs (that would most likely be the same in different node pool templates) somewhere, e.g. in the template context? This might help with the rate limits in clusters with a large number of node pools.

aermakov-zalando:
LGTM otherwise, but please use `aws.String/aws.StringValue` wrappers as @linki suggested.

provisioner/template.go
```diff
@@ -84,14 +89,15 @@ type podResources struct {
...
-func newTemplateContext(baseDir string, cluster *api.Cluster, nodePool *api.NodePool, values map[string]interface{}, userData string) *templateContext {
+func newTemplateContext(baseDir string, cluster *api.Cluster, nodePool *api.NodePool, values map[string]interface{}, userData string, adapter *awsAdapter) *templateContext {
 	return &templateContext{
 		baseDir:      baseDir,
 		manifestData: make(map[string]string),
 		cluster:      cluster,
 		nodePool:     nodePool,
 		values:       values,
 		userData:     userData,
+		awsAdpater:   adapter,
```
aermakov-zalando: Typo: _awsAdapter_.

aermakov-zalando:
👍

arjunrn:
:+1:

aermakov-zalando:
👍

PR #137, opened 2019-04-15
Check all err, fix ineffassign and structure files via goimports

njuettner:
* Use golangci-lint

PR #156, opened 2019-08-01
Checking Cluster Registry generated client

mikkeloscar:
This is in order to make the golangci-lint work which depends on all the files being present.

Follow up to #137


mikkeloscar:
:+1:

arjunrn:
We should also check the `Makefile` in this case so that the client is not generated on every build.

mikkeloscar:
> We should also check the `Makefile` in this case so that the client is not generated on every build.

I think the problem is that when it's checked out with git in the build system, the timestamp of the files might be different which causes make to thing it needs to regenerate. I'm not sure how to prevent this without loosing the dependency when we actually change something locally and wants make to pick it up?

linki:
:+1:

njuettner:
👍

njuettner:
@mikkeloscar 🙏 thanks!

PR #158, opened 2019-08-14
Improve err message after waiting for a CF Stack to be applied

muaazsaleem:
If there's an error after for a CF Stack to be applied, the StackStatus is sometimes too generic to understand the error message. Including the Reason in the error message so we don't have to look at the AWS console just for the error message.

Signed-off-by: Muhammad Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -293,7 +293,10 @@ func (a *awsAdapter) waitForStack(ctx context.Context, waitTime time.Duration, s
 			err = errUpdateRollbackFailed
 		}
 		if err != nil {
-			return fmt.Errorf("%s, reason: %s", err, *stack.StackStatusReason)
+			if stack.StackStatusReason != nil {
+				return fmt.Errorf("%s, reason: %s", err, *stack.StackStatusReason)
+			}
+			return err
 		}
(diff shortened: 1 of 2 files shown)


provisioner/aws.go
```diff
@@ -279,17 +280,23 @@ func (a *awsAdapter) waitForStack(ctx context.Context, waitTime time.Duration, s
...
 		case cloudformation.StackStatusUpdateRollbackComplete:
-			return errUpdateRollbackComplete
+			err = errUpdateRollbackComplete
 		case cloudformation.StackStatusUpdateRollbackFailed:
-			return errUpdateRollbackFailed
+			err = errUpdateRollbackFailed
+		}
+		if err != nil {
+			if stack.StackStatusReason != nil {
+				return fmt.Errorf("%s, reason: %s", err, *stack.StackStatusReason)
```
aryszka: I believe the format should be: "%v, reason: %s", because the value of `err` is not a string.

aryszka: the go vet command can detect these problems, may make sense to add it to the build

muaazsaleem: Thanks!

linki:
:+1:

muaazsaleem:
👍

PR #183, opened 2019-09-19
Directory config source: use an absolute path

aermakov-zalando:
This makes template context not fail if used with a non-absolute path like ".".


aermakov-zalando:
👍

aryszka:
:+1:

PR #192, opened 2019-10-15
nodeCIDRMaxPods: allow specifying extra capacity

aermakov-zalando:
This should allow us to still run master nodes with a CIDR of /27, because most of the pods there don't use any IPs.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -598,13 +598,13 @@ func nodeCIDRMaxNodes(maskSize int) (int, error) {
 	return 2 << (maskSize - 16 - 1), nil
 }
 
-func nodeCIDRMaxPods(maskSize int) (int, error) {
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

mikkeloscar:
I don't fully understand how this would be used, could you show an example in relation to Kubernetes-on-aws?

aermakov-zalando:
@mikkeloscar The idea is to add 5-10 extra pods to the master nodes because with 2048 nodes we'll have just 16 maximum pods which won't be enough for the masters. However, since most of the stuff running there uses host networking, they don't actually consume any IP addresses, which means we can increase the limit a bit.

mikkeloscar:
:+1:

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #193, opened 2019-10-17
InstanceInfo: drop storage device information

aermakov-zalando:
Storage is now automatically handled by the AMI, so we don't need this anymore. Dropping it in CLM allows us to mix instances with different storage in one spot pool, which can be useful for some clusters.


aermakov-zalando:
👍

aermakov-zalando:
👍

arjunrn:
:+1:

PR #190, opened 2019-10-14
Add template functions nodeCIDRMaxNodes, nodeCIDRMaxPods

aermakov-zalando:
We shouldn't have to calculate [these](https://github.com/zalando-incubator/kubernetes-on-aws/blob/1782ef98e6fcff5ff7cde56aaa76ccd45961d439/cluster/config-defaults.yaml#L259) manually.

PR #197, opened 2019-10-21
Template: add parseInt64

aermakov-zalando:
Follow-up for #190, #192. Easier than making all these functions parse as well.


linki:
:+1:

aermakov-zalando:
👍

PR #199, opened 2019-10-24
Fix autoscaling buffer settings

njuettner:
Selecting the right amount of CPU and Memory can get problematic when you have almost identical instance types.

This problem occured today and I'd like to address it with my PR.

```
error calling autoscalingBufferSettings: unable to select autoscaling buffer settings, conflicting instance types c4.4xlarge and c5.4xlarge
```

c4.4xlarge - 30.0 GiB | 16 vCPUs
c5.4xlarge - 32.0 GiB | 16 vCPUs

In this case CLM was not able to set the right settings for buffer pods.

We were able to identify this issue by adding more node pools in the test.

By changing a bit the logic of picking the CPU and Memory settings it won't have those problems.


provisioner/template.go
```diff
@@ -229,19 +229,23 @@ func autoscalingBufferSettings(clusterOrData interface{}) (*podResources, error)
...
 	}
 
+	if currentLargestVCPU == 0 || currentLargestMemory == 0 {
+		return nil, fmt.Errorf("unable to select autoscaling buffer settings, invalid CPU setting %d and Memory setting %d", currentLargestVCPU, currentLargestMemory)
+	}
+
 	result := &podResources{
-		CPU:    k8sresource.NewMilliQuantity(effectiveQuantity(currentLargestInstance.VCPU*1000, cpuScale, cpuReserved), k8sresource.DecimalSI).String(),
-		Memory: k8sresource.NewQuantity(effectiveQuantity(currentLargestInstance.Memory, memoryScale, memoryReserved), k8sresource.BinarySI).String(),
+		CPU:    k8sresource.NewMilliQuantity(effectiveQuantity(currentLargestVCPU*1000, cpuScale, cpuReserved), k8sresource.DecimalSI).String(),
```
aermakov-zalando: This should now compute the smallest available vCPU/memory instead, I think. The intent originally was to handle situations where you have node type combinations where some nodes have more CPU but less memory than the others. Previously this resulted in an error, and this code will configure autoscaling buffers that can't be scheduled at all.

aermakov-zalando:
👍

PR #201, opened 2019-10-29
Support multiple config sources

aermakov-zalando:
* Make Config an interface with accessors for the various manifests instead of a bare struct with a Path inside. This allows us to have Config implementations that can transparently merge multiple repositories, for example.
* Provisioner: use the new Config interface instead of manually mucking with the paths.
* Provisioner: drop CoreOS support.
* Provisioner: deprecate explicit `.UserData` and `.S3GeneratedFilesPath` template context values, users should update to `.Values.XXX` instead.
* Add a CombinedSource that pulls defaults, deletions and Kubernetes manifests from multiple sources.
* CLI: replace `--directory` and `--git-repository-url` with `--config-source`.


aermakov-zalando:
👍

channel/git.go
```diff
@@ -29,28 +29,27 @@ type Git struct {
 	mutex             *sync.Mutex
 }
 
-type GitVersions struct {
-	branches map[string]ConfigVersion
+type gitVersion struct {
+	git *Git
```
mikkeloscar: nit: this circular dependency on `Git` which is also the thing returning `gitVersion` is a little bit weird. But since it's all internal I'm ok with it :)

mikkeloscar:
:+1:

channel/config_source.go
```diff
@@ -6,28 +6,43 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-type ConfigVersion string
+type ConfigVersion interface {
+	// ID returns this version's unique ID (e.g. a git commit SHA)
+	ID() string
+
+	// Get returns a Config related to the specified version from the local copy.
```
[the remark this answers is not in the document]
aermakov-zalando: No, each `Config` is a self-contained clone (that you're supposed to `Delete()` later). `CachedConfig` just caches `ConfigVersions` to avoid unnecessary git invocations.

channel/git.go
aermakov-zalando: It's easier than copying most of the parts from the owning `Git`. I guess it could get an interface or something like a `func localClone(…)`, but I don't know if it'd improve the readability.

PR #202, opened 2019-11-01
template: add kmsEncrypt, kmsKeyARN functions

aermakov-zalando:
This is needed for configuring Scalyr for etcd instances.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -22,6 +22,7 @@ import (
 
 	awsUtil "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/aws"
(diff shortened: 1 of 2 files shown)


provisioner/template.go
```diff
@@ -280,6 +281,8 @@ func renderTemplate(context *templateContext, filePath string) (string, error) {
 		"nodeCIDRMaxNodes": nodeCIDRMaxNodes,
 		"nodeCIDRMaxPods":  nodeCIDRMaxPods,
 		"parseInt64":       parseInt64,
+		"kmsKeyARN":        func(keyID string) (string, error) { return kmsKeyARN(context.awsAdapter, keyID) },
```
mikkeloscar: For what do we need this function? Doesn't `kmsEncrypt` provide all we need or?

aermakov-zalando: You can't use an alias in an IAM policy, so we need a way to lookup a key using an alias. Technically `kmsEncrypt` returns both the ciphertext and the fully resolved key ARN, but I don't know if we should rely on it, plus it's going to look a bit magical in the templates.

mikkeloscar: Ok makes sense. Didn't realize we needed this for the IAM role definition.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #203, opened 2019-11-05
etcd: move scalyr management to AWS adapter

aermakov-zalando:
Turns out we don't use CLM templating at all, just `senza print`. This means that we need to add support for Scalyr stuff directly to CLM.


aermakov-zalando:
@mikkeloscar Sorry about that. :(

mikkeloscar:
@aermakov-zalando would it make sense to still keep the template functions and aim to move towards the template solution in the future?

I still like the idea of eventually getting rid of `senza` with a new improved etcd setup.

aermakov-zalando:
@mikkeloscar Sure, once we switch I'll just put it back into the templates. Or do you want to keep both?

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #210, opened 2019-11-22
Update to golangci-lint 1.21.0

mikkeloscar:
Update to golangci-lint 1.21.0

We can use a smaller build VM because the memory leak issue in golangci-lint has been fixed since the previous version. (was updated in the overlay)


njuettner:
👍

mikkeloscar:
:+1:

PR #211, opened 2019-11-23
Use make dependencies instead of calling make from make

mikkeloscar:
Use make dependencies instead of calling make from make


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #216, opened 2019-12-06
Rename spot_max_price profile to just spot

aermakov-zalando:
Easier to type, and allows us to tune the actual allocation strategy without having to rename every time.

### DIFF
--- a/api/node_pool.go
+++ b/api/node_pool.go
@@ -2,11 +2,6 @@ package api
 
 import "strings"
 
-const (
-	DiscountStrategyNone = "none"
-	DiscountStrategySpot = "spot_max_price"
-)
-
 // NodePool describes a node pool in a kubernetes cluster.
 type NodePool struct {
 	DiscountStrategy string            `json:"discount_strategy" yaml:"discount_strategy"`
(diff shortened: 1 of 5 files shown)


api/node_pool.go
```diff
@@ -21,6 +16,10 @@ type NodePool struct {
 	InstanceType string
 }
 
+func (np NodePool) IsSpot() bool {
+	return np.Profile == "spot_max_price" || np.Profile == "spot"
```
linki: `strings.Contains("spot")` maybe?

aermakov-zalando: This is just temporary so we could migrate to `spot`. Or do you think we'll have other spot profiles in the future?

aermakov-zalando: But yeah, we could just go with it. Fixed.

gargravarr: @linki I think if the need for more profiles shows up in the future, it can be addressed in another PR

linki: I see, didn't know it's a temporary thing. Fine for me like this as well.

aermakov-zalando:
👍

gargravarr:
:+1:

PR #221, opened 2019-12-11
Don't fail on missing cluster/manifests

aermakov-zalando:
There's no reason to fail if there are no manifests.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #230, opened 2020-01-06
All instances info

gargravarr:
This will allow for other projects to use the code handling AWS instance information. Ideally this should be decoupled from CLM - maybe in other open source repo: when are we having github.com/zalando-kubernetes? 😏


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #233, opened 2020-01-10
Make overriding a channel a bit easier

aermakov-zalando:
* `ConfigSource`: support multiple channels in `Version()`. The first one that exists in the source would be returned.
 * Prepend the value of `override_channel` config item to the channel list so we could use it for temporary experiments.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #245, opened 2020-01-28
OpenId configuration and provider

arjunrn:
This PR has the following changes:
- Provisioner for OpenID provider
- template functions to get openid discovery and JWKS document
- cleanup of `go.mod` and `go.sum`


provisioner/clusterpy.go
```diff
@@ -472,6 +498,20 @@ func (p *clusterpyProvisioner) Decommission(logger *log.Entry, cluster *api.Clus
...
 	}
 
+	// TODO: should this be done like this or via a config item?
+	hostedZone, err := getHostedZone(cluster.APIServerURL)
+	if err != nil {
+		return err
+	}
+
+	provisioner, err := p.oidcProvisioner(awsAdapter, cluster, hostedZone)
+	if err!=nil {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	if err != nil {
```

provisioner/jwks.go
```diff
@@ -0,0 +1,121 @@
+package provisioner
+
+import (
+	"crypto"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"fmt"

```

provisioner/openid_provider.go
```diff
@@ -0,0 +1,142 @@
+package provisioner
+
+import (
+	"crypto/sha1"
+	"crypto/x509"
+	"encoding/hex"
+	"encoding/pem"
+	"errors"
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"fmt"

```

provisioner/clusterpy.go
```diff
@@ -472,6 +498,20 @@ func (p *clusterpyProvisioner) Decommission(logger *log.Entry, cluster *api.Clus
...
+		return fmt.Errorf("failed to create oidc provisioner: %v", err)
+	}
+
+	logger.Infof("Decommissioning openid connect providers: %s (%s)", cluster.Alias, cluster.ID)
+	provisioner.Delete()
```
golangcibot: Error return value of `provisioner.Delete` is not checked (from `errcheck`)

provisioner/jwks.go
```diff
@@ -0,0 +1,121 @@
...
+
+// copied from kubernetes/kubernetes#78502
+func keyIDFromPublicKey(publicKey interface{}) (string, error) {
+	publicKeyDERBytes, err := x509.MarshalPKIXPublicKey(publicKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to serialize public key to DER format: %v", err)
+	}
+
+	hasher := crypto.SHA256.New()
+	hasher.Write(publicKeyDERBytes)
```
golangcibot: Error return value of `hasher.Write` is not checked (from `errcheck`)

provisioner/openid_provider.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		log.Infof("found provider: %s", *provider.Url)
+		if *provider.Url == p.providerUrl {
+			deletionCandidates = append(deletionCandidates, *prv.Arn)
+		}
+	}
+	for _, dc := range deletionCandidates {
+		_, err := p.awsAdapter.iamClient.DeleteOpenIDConnectProvider(&iam.DeleteOpenIDConnectProviderInput{OpenIDConnectProviderArn: &dc})
+		if err != nil {
+			fmt.Errorf("failed to delete openid connect provider with arn %s: %v", dc, err)
```
golangcibot: SA4017: Errorf is a pure function but its return value is ignored (from `staticcheck`)

provisioner/openid_provider.go
```diff
@@ -0,0 +1,147 @@
+package provisioner
+
+import (
+	"crypto/sha1"
+	"crypto/x509"
+	"encoding/hex"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	log "github.com/sirupsen/logrus"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

provisioner/openid_provider.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, dc := range deletionCandidates {
+		_, err := p.awsAdapter.iamClient.DeleteOpenIDConnectProvider(&iam.DeleteOpenIDConnectProviderInput{OpenIDConnectProviderArn: &dc})
+		if err != nil {
+			return fmt.Errorf("failed to delete openid connect provider with arn %s: %v", dc, err)
+		}
+	}
+	return nil
+}
+
+func stringPointerArray(input []string) []*string {
```
aermakov-zalando: AWS SDK has these: `aws.StringSlice()`/`aws.StringValueSlice()`.

provisioner/openid_provider.go
```diff
@@ -0,0 +1,178 @@
...
+	awsAdapter  *awsAdapter
+	providerUrl string
+	clientIDs   []string
+	thumbprints []string
+}
+
+func NewOpenIDProviderProvisioner(adapter *awsAdapter, clusterName, hostedZone string) (*OpenIDProviderProvisioner, error) {
+	return &OpenIDProviderProvisioner{
+		awsAdapter:  adapter,
+		providerUrl: fmt.Sprintf("%s.%s", clusterName, hostedZone),
```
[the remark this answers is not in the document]
arjunrn: The `APIServerURL` has the protocol as part of the value. I could use that and strip away the protocol but this is more straightforward IMO.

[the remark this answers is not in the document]
mikkeloscar: I don't agree that it is more straight forward to pass two parameters here which is only used concatenated. Why not concatenate/build the right format outside and pass a single parameter? Makes the API simpler IMO.

provisioner/jwks.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/pkg/errors"
+	"gopkg.in/square/go-jose.v2"
+)
+
+const (
+	kubernetesProgrammaticAuthorization = "urn:kubernetes:programmatic_authorization"
+	idTokenResponse                     = "id_token"
+	publicSubject                       = "public"
+	wellKnowOidcConfigurationUrl        = "https://%s.%s"
+	jwksUrl                             = "https://%s.%s/openid-configuration/keys.json"
```
szuecs: this URL you should get from `$issuer/.well-known/openid-configuration`

provisioner/jwks.go
```diff
@@ -0,0 +1,142 @@
...
+	default:
+		return "", errors.New("Public key was not RSA")
+	}
+
+	var alg jose.SignatureAlgorithm
+	switch pubKey.(type) {
+	case *rsa.PublicKey:
+		alg = jose.RS256
+	default:
+		return "", fmt.Errorf("invalid public key type %T, must be *rsa.PrivateKey", pubKey)
```
szuecs: this can't happen anymore, because in line 71-75 you already checked that

provisioner/jwks.go
```diff
@@ -0,0 +1,142 @@
...
+
+	kid, err := keyIDFromPublicKey(pubKey)
+	if err != nil {
+		return "", err
+	}
+	key := jose.JSONWebKey{
+		Key:       pubKey,
+		KeyID:     kid,
+		Algorithm: string(alg),
+		Use:       "sig",
```
szuecs: "sig" should be const

provisioner/openid_provider.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/iam"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	stsDomain = "sts.amazonaws.com"
+	// acmRootCAThumbprint is the SHA-1 sum of the root certificate in the trust chain for the certificate use to serve
+	// the open id discovery document.
+	acmRootCAThumbprint = "9e99a48a9960b14926bb7f3b02e22da2b0ab7280"
```
szuecs: How do we update this?
SHA-1 is already under attack by crypt analysis and I believe not a safe bet for long time anymore

arjunrn: As much as I would like to tell AWS that I don't think they will change what they accepts as [inputs for their API](https://github.bus.zalan.do/lago/s3-governance-control/blob/53d61110705af0e14c292ed179f6f53a21333bf6/roles/source-account-prod.yaml#L125).

provisioner/jwks.go
arjunrn: No, this is where I set the value for that field in the discovery document.

mikkeloscar: I think Sandors point is the same as mine, why not pass the full URL from APIServerURL here, instead of this injection which assumes a certain structure of the URL.

Similar to what I commented another place, but here you also need the protocol, so `APIServerURL` is truly the right value to use here IMO.

provisioner/openid_provider.go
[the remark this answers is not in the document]
arjunrn: fixed

provisioner/jwks.go
arjunrn: fixed

provisioner/jwks.go
arjunrn: fixed

mikkeloscar:
:+1:

provisioner/openid_provider.go
aermakov-zalando: Wrong link?

provisioner/openid_provider.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	var existingProvider *iam.GetOpenIDConnectProviderOutput
+	var existingProviderArn string
+	for _, prv := range providers.OpenIDConnectProviderList {
+		provider, err := p.awsAdapter.iamClient.GetOpenIDConnectProvider(&iam.GetOpenIDConnectProviderInput{OpenIDConnectProviderArn: prv.Arn})
+		if err != nil {
+			return fmt.Errorf("failed to get openid provider: %v", err)
+		}
+		if *provider.Url == p.providerUrl {
+			existingProvider = provider
```
aermakov-zalando: What happens if we accidentally have two providers with the same URL? Should we at least fail?

provisioner/openid_provider.go
```diff
@@ -0,0 +1,158 @@
...
+		if !stringArrayContains(aws.StringValueSlice(existingProvider.ThumbprintList), p.thumbprints) {
+			_, err := p.awsAdapter.iamClient.UpdateOpenIDConnectProviderThumbprint(&iam.UpdateOpenIDConnectProviderThumbprintInput{
+				OpenIDConnectProviderArn: aws.String(existingProviderArn),
+				ThumbprintList:           aws.StringSlice(mergeStringArrays(aws.StringValueSlice(existingProvider.ThumbprintList), p.thumbprints)),
+			})
+			if err != nil {
+				return fmt.Errorf("failed to update the thumprints for openid provider %s: %v", p.providerUrl, err)
+			}
+		}
+		if !stringArrayContains(aws.StringValueSlice(existingProvider.ClientIDList), p.clientIDs) {
```
aermakov-zalando: Do we really want to merge the new client IDs instead of just overwriting? What's the use case?

provisioner/openid_provider.go
arjunrn: AWS does [not allow](https://docs.aws.amazon.com/cli/latest/reference/iam/create-open-id-connect-provider.html#options) multiple providers with the same URL.

provisioner/openid_provider.go
```diff
@@ -0,0 +1,158 @@
...
+			return fmt.Errorf("failed to get openid provider: %v", err)
+		}
+		if *provider.Url == p.providerUrl {
+			existingProvider = provider
+			existingProviderArn = *prv.Arn
+		}
+	}
+
+	if existingProvider != nil {
+		if !stringArrayContains(aws.StringValueSlice(existingProvider.ThumbprintList), p.thumbprints) {
```
aermakov-zalando: Same question here: wouldn't it be easier to just overwrite the list with the expected value?

provisioner/openid_provider.go
```diff
@@ -0,0 +1,158 @@
...
+
+func makeStringMap(input []string) map[string]interface{} {
+	inputMap := make(map[string]interface{}, len(input))
+	for _, v := range input {
+		inputMap[v] = nil
+	}
+	return inputMap
+}
+
+func mergeStringArrays(first []string, second []string) []string {
```
aermakov-zalando: Something like `stringSetUnion` maybe so the purpose is more clear?

provisioner/openid_provider.go
arjunrn: Rotation maybe? Maybe we never have to update it but in case we have to due to an incident it would be easier this way.

provisioner/openid_provider.go
arjunrn: Again rotation of certificates

provisioner/openid_provider.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	var deletionCandidates []string
+
+	for _, prv := range providers.OpenIDConnectProviderList {
+		provider, err := p.awsAdapter.iamClient.GetOpenIDConnectProvider(&iam.GetOpenIDConnectProviderInput{OpenIDConnectProviderArn: prv.Arn})
+		if err != nil {
+			return fmt.Errorf("failed to get openid connect provider with arn %s: %v", *prv.Arn, err)
+		}
+		log.Infof("found provider: %s", *provider.Url)
```
aermakov-zalando: This looks more like `Debugf` to me.

provisioner/openid_provider.go
aermakov-zalando: I see!

arjunrn:
:+1:

provisioner/clusterpy.go
```diff
@@ -60,6 +56,7 @@ const (
 	updateStrategyCLC              = "clc"
 	defaultMaxRetryTime            = 5 * time.Minute
 	clcPollingInterval             = 10 * time.Second
+	serviceAccountPrivateKey       = "service_account_private_key"
```
aermakov-zalando: Where is this used?

arjunrn: unused :(

arjunrn:
:+1:

aermakov-zalando:
👍

provisioner/openid_provider.go
arjunrn: Fixed

PR #249, opened 2020-02-05
Add Stack.Outputs to Values and pass to templates

linki:
This captures all `Outputs` from Stack creation and update and passes them on to the templating engine where they can be used via the `Values` map. There's no conflict detection with other members of `Values` necessary as they follow a different naming scheme.

A practical use case can be found [here](https://github.com/zalando-incubator/kubernetes-on-aws/commit/65eedf5f838f0c74aec783b08bf5dad949867cb8#diff-bcfa1ed55a2ec099d9e371f395a6e151R439): A KMS key is created as part of the `cluster` Stack and returned via Outputs. Its value is then used as a flag in a Kubernetes sidecar.

This works well but if you have other ideas, please let me know.


provisioner/clusterpy.go
```diff
@@ -295,11 +295,15 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
...
 	bucketName := fmt.Sprintf(clmCFBucketPattern, strings.TrimPrefix(cluster.InfrastructureAccount, "aws:"), cluster.Region)
 
-	err = createOrUpdateClusterStack(ctx, channelConfig, cluster, values, awsAdapter, bucketName)
+	outputs, err := createOrUpdateClusterStack(ctx, channelConfig, cluster, values, awsAdapter, bucketName)
 	if err != nil {
 		return err
 	}
 
+	for _, o := range outputs {
+		values[aws.StringValue(o.OutputKey)] = aws.StringValue(o.OutputValue)
```
aermakov-zalando: Let's either have a prefix or put the outputs as a nested map (i.e. `values["ClusterStackOutputs"] = {…}`). We don't want to accidentally overwrite something important.

provisioner/clusterpy.go
```diff
@@ -365,29 +369,35 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
...
+		return nil, err
 	}
-	return nil
+
+	clusterStack, err := adapter.getStackByName(cluster.LocalID)
+	if err != nil {
+		return nil, err
+	}
+
+	return clusterStack.Outputs, nil
```
arjunrn: Maybe return a map of string instead over here.

provisioner/clusterpy.go
linki: They shouldn't clash as the naming scheme is different.

aermakov-zalando: Still, I'd group the stack outputs under one value (or with a prefix, but that's a bit worse in my opinion). It'll also be easier to figure out where something is coming from in the templates.

mikkeloscar:
You can also use kms alias to get a predictable name. Maybe you don't need this then? https://github.com/zalando-incubator/kubernetes-on-aws/blob/83a74d1d932c669786e701bbfa35c44597a64e3c/cluster/cluster.yaml#L708-L712

linki:
@mikkeloscar The docs [discourage](https://github.com/kubernetes-sigs/aws-encryption-provider#setup) the use of alias because one might be inclined to "move" the alias to a new key and expect a rotation, which will not work.

On the other hand we control all of this.

@mikkeloscar @arjunrn @aermakov-zalando Anyone prefers usage of Alias and dropping this feature?

arjunrn:
I think this feature is useful regardless. But prefixing it would be safer.

aermakov-zalando:
@linki It's not a lot of code, so if we can use key IDs I'd go for them.

provisioner/clusterpy.go
linki: Nice idea, changed it to a map.

provisioner/clusterpy.go
linki: Done, nested stack outputs under `values["ClusterStackOutputs"]`.

linki:
Alright, this is ready for another review then.

linki:
:+1:

aermakov-zalando:
👍

PR #252, opened 2020-02-11
fix CF stack deletion on cluster teardown 

szuecs:
we run regularly e2e tests an the cluster teardown is causing e2e to fail caused by ALB stacks failed to delete

Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -531,6 +531,9 @@ func (p *clusterpyProvisioner) Decommission(logger *log.Entry, cluster *api.Clus
 	for i := 0; i < maxApplyRetries; i++ {
 		// delete all cluster infrastructure stacks
 		err = p.deleteClusterStacks(ctx, awsAdapter, cluster)
+		if err == nil {
+			break
+		}
 	}
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

provisioner/clusterpy.go
```diff
@@ -521,17 +521,17 @@ func (p *clusterpyProvisioner) Decommission(logger *log.Entry, cluster *api.Clus
...
-	time.Sleep(time.Minute)
-
-	// we don't support cancelling decommission operations yet
-	ctx := context.Background()
-
-	// delete all cluster infrastructure stacks
-	err = p.deleteClusterStacks(ctx, awsAdapter, cluster)
+	for i := 0; i < maxApplyRetries; i++ {
+		// delete all cluster infrastructure stacks
+		err = p.deleteClusterStacks(ctx, awsAdapter, cluster)
```
aermakov-zalando: This could `break` if there are no stacks, but I guess doing 10 ListStacks is fast enough.

aermakov-zalando: This will take ages instead of failure (since `deleteClusterStack` retries internally on an error). `break` if there's an error since there's no point in retrying?
It could also `break` if there are no stacks, but I guess doing 10 `ListStacks` in a row is not a big deal.

szuecs: I put it into the condition

szuecs:
:+1:

aermakov-zalando:
👍

PR #257, opened 2020-02-19
Return a KB string for the kubernetesSizeToByte template function

arjunrn:
Prometheus needs the TSDB size to be specified in `KB`, `MB` or `GB`. So I have changed the return type of the `kubernetesSizeToBytes` to a string which returns _KB_s and change the name of the function accordingly.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #263, opened 2020-02-27
Make override_channel more reliable

aermakov-zalando:
Instead of silently falling back from `override_channel` to `channel` if the corresponding branch is missing, make it a hard failure. Additionally, change the format of `override_channel` so that the source that's being overridden is explicitly specified.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #272, opened 2020-03-09
Batch manifests when applying

aermakov-zalando:
We apply Kubernetes manifests with forked `kubectl apply`. This uses quite a lot of memory, so at some point we've set a limit on concurrent forks, and now our clusters take a very long time to update. Running `apply` in-process _might_ be feasible, but `kubectl` commands require a ton of complicated surroundings to setup, and the API is not guaranteed to be stable in any way, so it's unclear whether we should even try to go in this direction.
We can optimise the rollout speed a bit by batching the manifests before passing them to `apply`, which requires only very small changes to the code. For now let's batch all manifests for a module (i.e. `cluster/manifests/<foo>` directory), which makes us almost 4x faster while still keeping some sort of separation.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -71,6 +71,11 @@ type clusterpyProvisioner struct {
 	removeVolumes   bool
 }
 
+type manifestPackage struct {
+	name      string
+	manifests []string
+}
+
 // NewClusterpyProvisioner returns a new ClusterPy provisioner by passing its location and and IAM role to use.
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

aermakov-zalando:
Blocked by https://github.com/zalando-incubator/kubernetes-on-aws/pull/3054. I'd rather fix that instead of changing `UnmarshalStrict` to `Unmarshal`, because we shouldn't have issues like this in the first place.

PR #276, opened 2020-03-13
Use a truly unique directory for every checkout

aermakov-zalando:
The current approach (rely on `UnixNano`) isn't unique enough. Add a counter as well.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #285, opened 2020-04-06
Combined source: don't shadow components anymore

aermakov-zalando:
Shadowing was only useful for the initial migration, and now it just creates problems. Let's get rid of it.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #299, opened 2020-04-23
CLC: correctly delete node pools with non-0 min sizes

aermakov-zalando:
When we delete a pool, forcibly reset all ASGs to have min. size 0. Otherwise CLC will keep terminating nodes that AWS will replace with identical ones, so the node pool will never get deleted.


aermakov-zalando:
👍

muaazsaleem:
👍

PR #302, opened 2020-05-04
etcd: decode TLS-related config items before encrypting

aermakov-zalando:
We store them base64-encoded, so they need to be decoded first.

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -494,7 +494,11 @@ func (a *awsAdapter) CreateOrUpdateEtcdStack(parentCtx context.Context, stackNam
 	} {
 		if value, ok := cluster.ConfigItems[ci.configItem]; ok {
 			if ci.encrypt {
-				encrypted, err := a.kmsEncryptForTaupage(kmsKeyARN, value)
+				decoded, err := base64.StdEncoding.DecodeString(value)
+				if err != nil {
+					return err
+				}
+				encrypted, err := a.kmsEncryptForTaupage(kmsKeyARN, string(decoded))
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


provisioner/aws.go
```diff
@@ -494,7 +494,11 @@ func (a *awsAdapter) CreateOrUpdateEtcdStack(parentCtx context.Context, stackNam
 	} {
 		if value, ok := cluster.ConfigItems[ci.configItem]; ok {
 			if ci.encrypt {
-				encrypted, err := a.kmsEncryptForTaupage(kmsKeyARN, value)
+				decoded, err := base64.StdEncoding.DecodeString(value)
+				if err != nil {
+					return err
+				}
+				encrypted, err := a.kmsEncryptForTaupage(kmsKeyARN, string(decoded))
```
mikkeloscar, line 501: Won't we have the same problem for data which is not encrypted e.g. certificate data? This would be provided as base64 encoded PEM to the taupage config and thus to the environment variable and won't be decoded to PEM before being read by etcd as I understand our setup.

aermakov-zalando: Good point, thanks for noticing!

aermakov-zalando: Actually no, nevermind. All 3 config items containing certificates and keys will pass through this (they're all `encrypt: true`) because it's easier than dealing with newlines in Senza arguments, so I don't see any issues.

mikkeloscar:
:+1:

provisioner/aws.go
mikkeloscar: ok :+1: 

> because it's easier than dealing with newlines in Senza arguments

I assumed this was the reason for the base64 encoding in the first place :)

aermakov-zalando:
:+1:

PR #315, opened 2020-06-03
Create OIDC provider before the cluster

linki:
Creation of e2e clusters currently fail because the OIDC provider is created after the cluster. The provisioner is waiting for the cluster to be healthy which never happens and eventually times out.

This will also happen for new clusters once OIDC IAM is enabled by default with https://github.com/zalando-incubator/kubernetes-on-aws/pull/3296.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -309,6 +309,19 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

arjunrn:
@linki the provider in that case should also be decommissioned at the end.

linki:
@arjunrn Done, thanks.

arjunrn:
:+1:

mikkeloscar:
:+1:

PR #316, opened 2020-06-05
Add simple support for spot.io Ocean

mikkeloscar:
This introduces a simple way to support spot.io Ocean in CLM.

* Groups node pools by profile e.g. `master` vs. `worker` which allows updating in the right order. This additionally fixes an issue where master and worker configuration was applied at the same time such that new worker instances could be created e.g. with a new Kubernetes version before masters were updated. Now the order is:
  * Apply changes to master nodes and roll them
  * Apply changes to worker nodes and roll them (this includes spot.io nodes)
  * ~Apply changes to Spot IO workers (triggers a roll managed by Spot.io)~

UPDATE:

For spot.io nodes it has a special configuration for getting EC2 instance trough the EC2 api instead of via ASG which is done for the other pools.

Spot.io node pools has the profile `worker-spotio-ocean` and we map one spot.io Ocean to one node pool. It's possible to have multiple Ocean node pools in a cluster, but this will deploy the same amount of spot.io controllers.
Node pool update is managed by CLM + CLC and thus require that the update is disabled on the spot.io Ocean side.

~For the Spot.io part it will only be active the cluster has a node pool with the profile `worker-spotio-ocean` (And depends on this kubernetes-on-aws PR).~

This has a few limitations but could be used for PoC of spot.io:
* Deletion of spot.io node pools are not handled, so you have to delete the CF stack manually which may be disruptive to running pods.
* ~When updating the spot.io CF stack it can result in triggering a rolling update of the instances managed by spot.io. This rolling update is not clear from the cloudformation API as the stack would just be in a `COMPLETE` state, however if re-applying the stack while a rolling update is in progress the stack can end up in `ROLLBACK_FAILED` state and have to me manually recovered. The only way to check the rolling state is via spot.io API (Not implemented in CLM for now).~


pkg/spotio/spotio.go
```diff
@@ -0,0 +1,127 @@
...
+// API is an interface describing the Spot.io HTTP API.
+type API interface {
+	ListLaunchSpecs(oceanID string) ([]OceanLaunchSpec, error)
+}
+
+// Client is the implementation of a spot.io HTTP client.
+type Client struct {
+	baseURL   string
+	accountID string
+	client    *http.Client
```
linki, line 44: Looks like we could also use their golang-sdk which might make it easier and safer to call their API.

https://github.com/spotinst/spotinst-sdk-go/blob/master/service/ocean/providers/aws/service.go#L14

linki:
Looks understandable to me.

pkg/spotio/spotio.go
mikkeloscar: Good point. I didn't even check if they had one, but seems easy to use and the http client have timeouts etc. :) I will try this one.

pkg/updatestrategy/aws_asg.go
```diff
@@ -25,7 +25,8 @@ const (
 	clusterIDTagPrefix               = "kubernetes.io/cluster/"
 	resourceLifecycleOwned           = "owned"
 	kubeAutoScalerEnabledTagKey      = "k8s.io/cluster-autoscaler/enabled"
-	nodePoolTag                      = "NodePool"
+	nodePoolTagLegacy                = "NodePool"
```
aermakov-zalando, line 28: Do we even use it?

pkg/updatestrategy/aws_ec2.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/spotio"
+)
+
+const (
+	spotIOOceanIDTagKey = "spot.io/ocean-id"
+)
+
+// EC2NodePoolsBackend defines a node pool consiting of externally managed EC2
+// instances
+type EC2NodePoolsBackend struct {
```
aermakov-zalando, line 25: `SpotIOBackend`? Or do you see other potential instance managers?

provisioner/clusterpy.go
```diff
@@ -1199,3 +1225,38 @@ func int32Value(v *int32) int32 {
 	}
 	return 0
 }
+
+func groupNodePools(logger *log.Entry, cluster *api.Cluster, updater updatestrategy.UpdateStrategy) []nodePoolGroup {
+	var masters, workers []*api.NodePool
+	for _, nodePool := range cluster.NodePools {
+		if strings.HasPrefix(nodePool.Profile, "master") {
```
aermakov-zalando, line 1232: `strings.Contains`? A bit less error-prone.

provisioner/clusterpy.go
```diff
@@ -1199,3 +1225,38 @@ func int32Value(v *int32) int32 {
...
+			ReadyFn: func() error {
+				return nil
+			},
+		},
+	}
+}
+
+type nodePoolGroup struct {
+	NodePools []*api.NodePool
+	Updater   updatestrategy.UpdateStrategy
```
aermakov-zalando, line 1260: We only have one right now (I just didn't have time to get rid of it), is it really needed?

aermakov-zalando:
Mostly LGTM, but let's think about refactoring it a bit later!

provisioner/clusterpy.go
mikkeloscar: Changed

provisioner/clusterpy.go
mikkeloscar: Good point, dropped it from here. Initially I was implementing a different (no-op) updater but this is now changed so no need to carry it per node pool group.

pkg/updatestrategy/aws_asg.go
mikkeloscar: This is still the tag used in `aws_asg.go` to find the right asg for a node pool. I would like to change this but will do so separately.

pkg/updatestrategy/aws_ec2.go
mikkeloscar: The idea was to have a generic EC2 pool and then use whatever way to find the desired config. Right now I explicitly use the spotio API for this, but I wanted to abstract this part. However for now we don't have other use cases, so I can make it spot.io specific.

mikkeloscar: Changed!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #317, opened 2020-06-22
Add configItem to mark nodes with NoSchedule taint on decommissioning

mikkeloscar:
Adds a config item `clm_decommission_node_no_schedule_taint` which when set to `true` will make CLM mark nodes intended for decommissioning with a `NoSchedule` taint. If the configItem is not set the taint will still be added but with `PreferNoSchedule` as it is now.

The motivation for this change is to ensure that pods being drained during a cluster update will not land back on old nodes and have to be drained again.
The intention is to enable this behavior by default as we expect our current cluster autoscaling setup  is fast enough to provide new nodes when needed (i.e. our Pod schedule SLO).

It's implemented as a config item so we can quickly disable it in case it doesn't have the desired effect.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -35,30 +35,31 @@ import (
 )
 
 const (
-	providerID                     = "zalando-aws"
-	etcdStackFileName              = "etcd-cluster.yaml"
-	clusterStackFileName           = "cluster.yaml"
-	defaultNamespace               = "default"
-	kubectlNotFound                = "(NotFound)"
-	tagNameKubernetesClusterPrefix = "kubernetes.io/cluster/"
-	subnetELBRoleTagName           = "kubernetes.io/role/elb"
-	resourceLifecycleShared        = "shared"
-	resourceLifecycleOwned         = "owned"
-	mainStackTagKey                = "cluster-lifecycle-controller.zalando.org/main-stack"
-	stackTagValueTrue              = "true"
-	subnetsConfigItemKey           = "subnets"
-	subnetsValueKey                = "subnets"
-	availabilityZonesConfigItemKey = "availability_zones"
-	availabilityZonesValueKey      = "availability_zones"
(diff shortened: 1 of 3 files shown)


provisioner/clusterpy.go
```diff
@@ -748,8 +749,13 @@ func (p *clusterpyProvisioner) prepareProvision(logger *log.Entry, cluster *api.
 		}
 	}
 
+	noScheduleTaint := false
+	if v, ok := cluster.ConfigItems[decommissionNodeNoScheduleTaintKey]; ok && v == "true" {
```
aermakov-zalando, line 753: You could just test for `v == "true"` here.

provisioner/clusterpy.go
```diff
@@ -35,30 +35,31 @@ import (
...
+	vpcIDConfigItemKey                 = "vpc_id"
+	subnetAllAZName                    = "*"
+	maxApplyRetries                    = 10
+	configKeyUpdateStrategy            = "update_strategy"
+	updateStrategyRolling              = "rolling"
+	updateStrategyCLC                  = "clc"
+	defaultMaxRetryTime                = 5 * time.Minute
+	clcPollingInterval                 = 10 * time.Second
+	clusterStackOutputKey              = "ClusterStackOutputs"
+	decommissionNodeNoScheduleTaintKey = "clm_decommission_node_no_schedule_taint"
```
aermakov-zalando, line 62: Let's drop the `clm_` part? It shouldn't matter which application does that, and if we ever move it to CLC it'll just be more confusing.

mikkeloscar: done

provisioner/clusterpy.go
mikkeloscar: done

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #318, opened 2020-06-25
Use client-go instead of forking out to kubectl to handle deletions

aermakov-zalando:
This should make our updates slightly faster (and use less resources) since we won't need to constantly spawn kubectls just to delete things from the cluster. The code to resolve Kind into a GVR was adapted from [cli-runtime](https://github.com/kubernetes/cli-runtime/blob/be96dd2fb350ca2e227f26978d7cafdc56399636/pkg/resource/builder.go#L695), it seems to work fine for most of our use cases but we can change it later if we want fancier things.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -34,7 +34,13 @@ import (
 	"github.com/zalando-incubator/kube-ingress-aws-controller/certs"
 	"golang.org/x/oauth2"
 	"gopkg.in/yaml.v2"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/restmapper"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


provisioner/clusterpy.go
```diff
@@ -976,52 +981,111 @@ type deletions struct {
...
+}
+
+func processDeletion(client dynamic.Interface, mapper meta.RESTMapper, logger *log.Entry, deletion *resource) error {
+	// Figure out the GVR
+	gvr, err := resolveKind(mapper, deletion.Kind)
+	if err != nil {
+		return err
+	}
+
+	// indentify the resource to be deleted either by name or
```
mikkeloscar, line 1055: nit: indentiy -> identify

provisioner/clusterpy.go
```diff
@@ -976,52 +981,111 @@ type deletions struct {
...
+		gvr, _ = mapper.ResourceFor(groupResource.WithVersion(""))
+	}
+	if gvr.Empty() {
+		return schema.GroupVersionResource{}, fmt.Errorf("unable to resolve kind %s (use either name or name.version.group)", kind)
+	}
+	return gvr, nil
+}
+
 // Deletions uses kubectl delete to delete the provided kubernetes resources.
-func (p *clusterpyProvisioner) Deletions(ctx context.Context, logger *log.Entry, cluster *api.Cluster, deletions []*resource) error {
```
mikkeloscar, line 980: No longer uses `kubectl` :)

provisioner/clusterpy.go
aermakov-zalando: Fixed!

provisioner/clusterpy.go
aermakov-zalando: Fixed!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #329, opened 2020-07-10
Drop autoscalingBufferSettings and related code

aermakov-zalando:
They're way too complicated to use, and we don't even rely on them that much anymore. Switching to static sizes is easier.

**IMPORATANT**: don't merge until the [kubernetes-on-aws PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/3402) is fully rolled out.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -13,8 +13,6 @@ import (
 	"net"
 	"net/url"
 	"path"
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

aryszka:
:+1:

PR #357, opened 2020-09-07
Add support for Metadata.Tags in CF stacks

mikkeloscar:
This adds support for defining cloud formation Tags via the cloudformation yaml template similar to what we have for our internal CDP deployments. Basically it will be possible to add:

```yaml
Metadata:
  Tags:
    foo: bar
```

To a template in kubernetes-on-aws and the tags will be added when the stack is being created. This allows setting custom tags like `InfrastructureComponent=true` which we can use to control permissions.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #358, opened 2020-09-17
Use a non-cancelled context for aborting the update

aermakov-zalando:
The current implementation just doesn't work at all now that client-go actually uses the context.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #359, opened 2020-09-18
Update kubectl used in docker image

mikkeloscar:
Update the `kubectl` binary used in CLM to match with the cluster version.


mikkeloscar:
:+1:

jrake-revelant:
👍

PR #360, opened 2020-09-24
allow indexed lists in templates

aryszka:
allow indexed lists in templates

Signed-off-by: Arpad Ryszka <arpad.ryszka@gmail.com>

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -649,13 +649,13 @@ func extractEndpointHosts(endpoints string) ([]string, error) {
 	return result, nil
 }
 
-func indexedList(itemTemplate string, length int) (string, error) {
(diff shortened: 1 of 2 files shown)


provisioner/template.go
```diff
@@ -647,3 +648,16 @@ func extractEndpointHosts(endpoints string) ([]string, error) {
 	sort.Strings(result)
 	return result, nil
 }
+
+func indexedList(itemTemplate string, length int) (string, error) {
```
aermakov-zalando, line 652: The `length` argument will come from the config items and all the values there are strings. You want to either accept a string argument here, and do the conversion inside, or change it to `int64` because we already have a `parseInt64` function.

aermakov-zalando:
👍

aryszka:
:+1:

PR #371, opened 2020-10-01
Update AWS instance types

aermakov-zalando:
Add `c5ad`, `t4g` instances. I've had to patch storage information manually for a couple of those, maybe we should switch away from ec2instances.info.


aermakov-zalando:
👍

mikkeloscar:
> maybe we should switch away from ec2instances.info.

Or contribute the information? :) Unless there is a better source nowadays?

mikkeloscar:
:+1:

aermakov-zalando:
@mikkeloscar AWS has some kind of an API for this now, which is used in the CA generator for example.

PR #372, opened 2020-10-05
Correctly handle cluster-scoped deletions

aermakov-zalando:
The dynamic client apparently doesn't resolve the scope of the resource on its own. This means that we have to do it ourselves, which is rather cumbersome. A much simpler solution is dropping the defaulting in the `deletions.yaml` parser, which we don't really use anyway, and then relying on the presence/absence of the namespace field instead.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #382, opened 2020-11-09
Add zoneDistributedNodePoolGroups template function

aermakov-zalando:
`zoneDistributedNodePoolGroups ` returns information on whether a node pool group (same value of the 'dedicated' label) is are safe to use with even pod spreading. Currently this is the case iff all node pools with this dedicated label
  - are correctly configured with regards to the labels and taints
  - don't have AZ restrictions
  - use the worker-splitaz profile.

The default node pool group is represented by the empty string, and it additionally ignores individual node pools with taints.


mikkeloscar:
:+1:

aermakov-zalando:
👍

linki:
:+1:

PR #401, opened 2020-12-11
Support non-ocean spotio node pool

mikkeloscar:
Defined a new node pool profile: `worker-spotio` where the spotio setup is different thant `worker-spotio-ocean` profiles, but from CLM PoV it needs to be handled the same way.

This is needed for https://github.com/zalando-incubator/kubernetes-on-aws/pull/3787 and later we can drop support from `worker-spotio-ocean` when we have migrated away.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #408, opened 2021-01-08
Fetch instance information from AWS

aermakov-zalando:
Changes: 
 * Get rid of the static `instances.json` file and fetch instance type information from AWS when provisioning a cluster. This means that we don't need to do anything when AWS adds a new instance type.
 * Drop `NodePool.AvailableStorage` completely.

**Note**: this can only be merged once #409 has been merged and all cluster configurations have migrated to `.Values.InstanceInfo.AvailableStorage`.


aermakov-zalando:
👍

mikkeloscar:
:+1:

aermakov-zalando:
Apparently the old code parsed the storage size incorrectly (using GiB instead of GB), so this will roll all the nodes with local storage. Let's hold it until we're ready.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #409, opened 2021-01-08
Deprecate NodePool.AvailableStorage, move to instance info instead

aermakov-zalando:
This is a preparation for #408.

Changes:
 * Rename `.Values.instance_info` -> `.Values.InstanceInfo`. It wasn't used anywhere so it's safe.
 * Deprecate `.NodePool.AvailableStorage` (since it relies on global AWS data which will not be readily available. Add `.Values.InstanceInfo.AvailableStorage` instead.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #410, opened 2021-01-13
Wrap spot.io calls in backoff retry logic

jonathanbeber:
Spot.io has an untreated error message returned on their API and it's
generating [`problems`][0] on clusters frequently. The error is
transient and works with a retry. We opened a support ticket with them.

Until a proper solution, this code wraps the API call in a backoff retry
logic of max 5 retries with 1 second of constant backoff.

[0]: https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/8e53ebb8fb1e52488e439e5352f9c3df58ab4bd4/api/cluster_status.go#L8

Signed-off-by: Jonathan Juares Beber <jonathanbeber@gmail.com>

### DIFF
--- a/pkg/updatestrategy/aws_spotio.go
+++ b/pkg/updatestrategy/aws_spotio.go
@@ -3,6 +3,7 @@ package updatestrategy
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #411, opened 2021-01-19
Debug when retrying deployments downscale

jonathanbeber:
When decommissioning clusters where API server is not running, the
operation might seem stuck, even with `--debug` since the retries of
deployments downscale are running in silent. It will run after the
backoff retries, but adding a debug message make it quicker to
understand.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -524,7 +524,11 @@ func (p *clusterpyProvisioner) Decommission(ctx context.Context, logger *log.Ent
 	// recreate resources we delete in the next step
 	err = backoff.Retry(
 		func() error {
-			return p.downscaleDeployments(ctx, logger, cluster, "kube-system")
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
Do we even need the scaledown? We'll delete everything afterwards, so I'm not sure if I see the point.

jonathanbeber:
@aermakov-zalando good point. I think I could check the code a bit further. I don't see any reason to care about it too.

mikkeloscar:
Scaling down is a hack to stop any deployments which may be creating stacks like ingress-controller and egress-controller. Without it we run into a situation where an ingress stack is being deleted and then a new one is created by the controller before the cluster has been completely terminated.

jonathanbeber:
thanks, @mikkeloscar. So I think we can consider this PR as it's.

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #412, opened 2021-01-19
AvailableStorage: return int64, add tests

aermakov-zalando:
Follow-up to #409, fixes the return type of `NodePool.AvailableStorage`.


aermakov-zalando:
👍

PR #414, opened 2021-01-25
Add additional OIDC providers from config item (POC)

linki:
This allows to specify additional allowed OIDC providers via config item.

If a cluster A wants to allow cross-account access from clusters B and C you would set the following config item in cluster A:

```
oidc_additional_providers: B, C
```

Biggest flaw of this approach: I don't see a good way to remove OIDC providers if items are removed from the list as well as in multi-cluster accounts, such as e2e.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -72,6 +72,7 @@ const (
 	spotIONodePoolProfileLegacy        = "worker-spotio-ocean"
 	spotIONodePoolProfile              = "worker-spotio"
 	decommissionNodeNoScheduleTaintKey = "decommission_node_no_schedule_taint"
+	oidcAdditionalProvidersConfigItem  = "oidc_additional_providers"
 )
 
 type clusterpyProvisioner struct {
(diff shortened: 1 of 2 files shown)


mikkeloscar:
> Biggest flaw of this approach: I don't see a good way to remove OIDC providers if items are removed from the list as well as in multi-cluster accounts, such as e2e.

Is it possible to tag OIDC providers? If so we could tag them with the cluster ID as we normally do for other resources. Then it's possible with a bit more logic to clean them up.

linki:
I think there's no support for tags :disappointed: 

```console
$ aws iam get-open-id-connect-provider --open-id-connect-provider-arn arn:aws:iam::170858875137:oidc-provider/kube-aws-test-1.teapot.zalan.do                                       ✗ ✭ ✱
{
    "Url": "kube-aws-test-1.teapot.zalan.do",
    "ClientIDList": [
        "sts.amazonaws.com"
    ],
    "ThumbprintList": [
        "9e99a48a9960b14926bb7f3b02e22da2b0ab7280"
    ],
    "CreateDate": "2020-02-06T09:17:35.597Z"
}
```

mikkeloscar:
@linki this we can close right?

linki:
Yes, this can be closed.

PR #427, opened 2021-02-11
Improve userdata files

aermakov-zalando:
* Use the correct KMS key depending on the pool type
 * Include the cluster ID and the pool kind so it's easier to cleanup later (only if `clm_new_userdata_path` is set to true)

### DIFF
--- a/api/node_pool.go
+++ b/api/node_pool.go
@@ -1,6 +1,8 @@
 package api
 
 import (
+	"strings"
+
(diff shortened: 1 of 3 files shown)


aermakov-zalando:
👍

aermakov-zalando:
Argh, wait, this will roll all the nodes.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #429, opened 2021-02-12
Add support for sgIngressRanges

mikkeloscar:
Adds a new template function `sgIngressRanges` which is similar to `portRanges` but allows specifying optional CIDR in the ranges.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -109,6 +109,7 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"mountUnitName":        mountUnitName,
 		"accountID":            accountID,
 		"portRanges":           portRanges,
+		"sgIngressRanges":      sgIngressRanges,
 		"splitHostPort":        splitHostPort,
 		"extractEndpointHosts": extractEndpointHosts,
 		"publicKey":            publicKey,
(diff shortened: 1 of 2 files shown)


provisioner/template.go
```diff
@@ -314,6 +315,62 @@ func portRanges(ranges string) ([]PortRange, error) {
...
+type SGIngressRange struct {
+	CIDR             string
+	FromPort, ToPort int
+}
+
+// sgIngressRanges parses a list of Security Group Ingress ranges.
+//
+// Example:
+//
+// "10.0.0.0/8:4180-4180,4190-4190" would result in the ingress ranges:
```
jonathanbeber, line 327: I found it a bit confusing, we could always require the `:` even and fallback to `0.0.0.0/0` when it's empty, for example: `"10.0.0.0/8:4180-4180,:4190-4190"` (just like when you run an HTTP server but doesn't want to specify the IP `my_server run --listen=:8080`)

I expected `"10.0.0.0/8:4180-4180,4190-4190"` to be `[{CIDR: "10.0.0.0/8", FromPort: 4180, ToPort: 4180},{CIDR: "10.0.0.0/8", FromPort: 4190, ToPort: 4190}]` (same CIDR repeated to both port ranges) .

Not a hard feeling, but it might be confusing. We could also never fallback to `0.0.0.0/0` and always make it explicit. WDYT?

szuecs: I think we discussed in the video call to do:
```
"10.0.0.0/8:4180-4180,10.0.0.0/8:4190-4190"
```

So port range needs always be prepended by CIDR.
Like this you can have 2 different port ranges for 2 different CIDRs and in code it would be easy to map, not much parsing needed.

aermakov-zalando: That's a good point, being explicit about the source IP sounds better and less confusing to me.

Considering that we only have the corresponding config item (`open_port_ranges`) set in 4 clusters, it might be easier to just adjust the existing `portRanges` function to require the source CIDR, merge the CLM changes and then just update the config item afterwards. This is a bit more hacky, but a lot easier than a full-blown migration, and shouldn't cause any issues other than a possible failed update or two. WDYT @mikkeloscar?

mikkeloscar: The original idea was to be backwards compatible with the old format, but I agree that having it explicit is more clear and it's not so much effort to migrate. I have changed it as described in the comment:

```
"10.0.0.0/8:4180-4181,0.0.0.0/0:4190" would result in the ingress ranges:
[{CIDR: "10.0.0.0/8", FromPort: 4180, ToPort: 4181},{CIDR: "0.0.0.0/0", FromPort: 4190, ToPort: 4190}]
```

Not I also made the `-` optional so if from and to port is the same you don't need to specify `1111-1111` but can just say `1111`. I think this is simpler.

Regarding re-use of `portRanges` I think we eventually should migrate to the other name as it's IMO confusing in the long run if it includes CIDRs but is just called `portRanges`. But we can ofc. start with re-using the function and then later changes the function and config item name. WDYT?

jonathanbeber:
:+1:

provisioner/template.go
aermakov-zalando: SGTM, let's migrate correctly!

mikkeloscar:
:+1:

mikkeloscar:
:+1:

PR #432, opened 2021-02-19
Add a template function to check for spot.io pools

mikkeloscar:
Adds a template function which returns true if any of the node pools is a spot.io pool. This will make it simple to enable/disable spot.io components based whether it's used or not.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #444, opened 2021-03-05
etcd stack: pass certificate expiry information 

aermakov-zalando:
We use these for monitoring, so we should set this info when we create the stack.

**Warning**: will block new cluster creation unless https://github.com/zalando-incubator/kubernetes-on-aws/pull/4121 is in the corresponding channel.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #445, opened 2021-03-05
Fix worker-spotio support

mikkeloscar:
* Fix supporting multiple node pools of type `worker-spotio` by looking up the right LaunchSpec based on node pool name
* Drop support for `worker-spotio-ocean` (one ocean per node pool setup). This is no longer compatible with the above change AND it's no longer used.


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #446, opened 2021-03-22
Extend sgIngressRanges to support optional protocol

mikkeloscar:
Extends the `sgIngressRanges` template function to also support an optional protocol `(tcp|udp)`. If no protocol it falls back to `tcp` which was the only one supported before.

The new format is `[<protocol>:]<cidr>:<port>[-<port>],`

The reason for introducing `udp` support is that UDP NLBs can now be deployed since Kubernetes v1.19


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #457, opened 2021-04-01
Update dependencies

jonathanbeber:
This commit is the output of `go get -u` and `go mod tidy`.


mikkeloscar:
We may want to exclude the client-go (+friends) v1.20.x update until we run v1.20. We can do v1.19.9 for now.

jonathanbeber:
Good point, I will revert these

jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #459, opened 2021-04-06
Drop support for clm_new_userdata_path

aermakov-zalando:
The [kubernetes-on-aws PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/4031) has been rolled out everywhere, so we don't need this anymore.

Note that this still doesn't implement any kind of cleanup, but we can at least drop the config item.


aermakov-zalando:
👍

jrake-revelant:
👍

PR #460, opened 2021-04-21
Always use PreferNoSchedule for master nodes

mikkeloscar:
When `noScheduleTaint=true` nodes about to be decommissioned will get a `NoSchedule` taint to avoid new pods being scheduled there.
For master nodes it's possible to end in a situation where two separate master node updates can cause all current nodes to be marked for decommissioning meaning nothing new can be scheduled on them. When we do a rolling update we want the Cluster Lifecycle Controller (CLC) (and other master pods) to be moved to one of the other nodes when the current nodes is being decommissioned. When all nodes are marked for decommissioning the pods can no longer be scheduled resulting in the update not continuing.

This changes the taint such that for `master` nodes it's always `PreferNoSchedule`. This can have the side effect that master pods are bounced a bit around between the nodes, but it's better than having them not running at all.

### DIFF
--- a/pkg/updatestrategy/node_pool_manager.go
+++ b/pkg/updatestrategy/node_pool_manager.go
@@ -118,6 +118,8 @@ func (m *KubernetesNodePoolManager) GetPool(ctx context.Context, nodePoolDesc *a
 
 	nodes := make([]*Node, 0, len(instanceIDMap))
 
+	master := nodePoolDesc.IsMaster()
+
 	for _, npNode := range nodePool.Nodes {
 		if node, ok := instanceIDMap[npNode.ProviderID]; ok {
 			n := &Node{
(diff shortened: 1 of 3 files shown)


pkg/updatestrategy/node_pool_manager.go
```diff
@@ -120,6 +121,10 @@ func (m *KubernetesNodePoolManager) GetPool(ctx context.Context, nodePoolDesc *a
 
 	for _, npNode := range nodePool.Nodes {
 		if node, ok := instanceIDMap[npNode.ProviderID]; ok {
+			role := WorkerNodeRole
+			if r, ok := node.Annotations[nodeRoleAnnotaton]; ok && NodeRole(r) == MasterNodeRole {
```
aermakov-zalando, line 125: We have `NodePool.IsMaster()`, could we use that one instead?

mikkeloscar: Changed to decide node type based on the node pool

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #461, opened 2021-04-27
When only 1 VPC, assume default.

gargravarr:
To avoid depending on a VPC being the default, but only when only 1 VPC exists in the account. 


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #462, opened 2021-04-27
Add test for Default VPC

gargravarr:
Since I was already coding a test, let's increase code coverage.


gargravarr:
:+1:

provisioner/aws_test.go
```diff
@@ -251,6 +267,14 @@ func TestGetStackByName(t *testing.T) {
 	}
 }
 
+func TestGetDefaultVPC(t *testing.T) {
+	a := newAWSAdapterWithStubs("", "GroupName")
+	_, err := a.GetDefaultVPC()
```
mikkeloscar, line 272: maybe also check that the vpc returned here is not `nil`.

gargravarr:
:+1:

mikkeloscar:
:+1:

PR #481, opened 2021-06-10
Manage the etcd stack with CLM

aermakov-zalando:
The current state of our etcd stack is, uh, _less than ideal_. We still rely on `senza` to create the CloudFormation stack in the first place, and then never update it directly. Instead, we rely on a bunch of hacky scripts that run on our laptops and use ZMON as a data source. We also still rely on the Taupage AMI which has been deprecated and EOL'd a long time ago.

Let's slowly start migrating towards a more sane setup. This PR implements the logic to actually update the etcd CloudFormation stack with CLM. It's kinda hacky because we still have to be compatible with Taupage, but this would allow us to 1) drop `senza` completely and 2) implement the etcd node rotation as a cronjob running in the cluster without jumping through a bunch of hoops.

The new behaviour is enabled with a dedicated config item (`experimental_new_etcd_stack`), so this should be fairly safe to merge. In addition to that, management of etcd stacks in general (both creation and updates) is now disabled by default, and only enabled with a dedicated command-line argument (`--manage-etcd-stack`). This is mainly to allow for easy provisioning of pet clusters without running into tons of conflicts.

### DIFF
--- a/provisioner/legacy_etcd.go
+++ b/provisioner/legacy_etcd.go
@@ -1,0 +1,115 @@
+package provisioner
+
+import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// populateEncryptedEtcdValues is a horrible hack that is needed to avoid doing an update of the Taupage-based CloudFormation stack every time
+// we run. Unfortunately it's pretty much unavoidable because we rely on the KMS-encrypted config items inside the launch template, and if we
+// didn't do this, every run would create new ciphertext for the same plaintext value and then trigger a rolling update. We can drop it and switch
+// to a saner scheme (same as what we do with Kubernetes) once we migrate away from Taupage, and then this whole mess can be dropped.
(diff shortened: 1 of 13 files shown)


provisioner/legacy_etcd.go
```diff
@@ -0,0 +1,115 @@
...
+func populateEncryptedEtcdValues(adapter *awsAdapter, cluster *api.Cluster, etcdKMSKeyARN string, values map[string]interface{}) error {
+	stack, err := adapter.getStackByName(etcdStackName)
+	if err != nil && !isDoesNotExistsErr(err) {
+		return err
+	}
+
+	// Try to reload the values from the existing launch template
+	if stack != nil {
+		for _, output := range stack.Outputs {
+			if aws.StringValue(output.OutputKey) == "LaunchTemplateId" {
```
mikkeloscar, line 29: Is this the right name and not `etcd-cluster-etcd:launch-template-id` which is defined as the export name in the stack?

aermakov-zalando: The export name is stored in `output.ExportName`. Since it's not really used anywhere else, I've decided to use the key in CLM, but I don't really have a strong preference here. Should I change it to use the export name?

mikkeloscar: No, it's fine, thanks for clarifying!

mikkeloscar:
:+1:

aermakov-zalando:
👍

gargravarr:
After reading the PRs and issue description, I would e to test the following scenarios in Teapot cluster:

1. Set the Config Item `experimental_new_etcd_stack` for teapot cluster
2. "Upgrade" etcd using the upgrade script (make the necessary changes)

or/and

1. Set the Config Item `experimental_new_etcd_stack` for teapot cluster
2. Decommission the cluster
3. Recommission the cluster.

PR #490, opened 2021-07-08
Drop support for spot.io node pools

mikkeloscar:
Drop all spot.io support as we will no longer use it.

Ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/4392


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #501, opened 2021-08-20
Fix the etcd value parsing for the new stack template

aermakov-zalando:
We have to handle both base64-encoded values and the raw ones, and it's easier to just add a small hack since this whole thing will be dropped afterwards.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #512, opened 2021-09-03
Update kubectl to a more recent version

aermakov-zalando:
It looks like we forgot to do this when we've updated `go.mod`.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #513, opened 2021-09-07
etcd stack: update the application/component tags

aermakov-zalando:
Let's create new stacks with the correct application name, I'll update the existing ones manually. Also fixes the tags on the experimental senza-less setup.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #543, opened 2021-12-20
[do not merge] Print git status to debug version tag

AlexanderYastrebov:
Version tag is generated based on the `git describe --tags --always --dirty`.
Starting from `9a6ffb4-dirty` it has `-dirty` suffix signalling that
working tree has local modification.


AlexanderYastrebov:
Superseded by #544

PR #544, opened 2021-12-20
Print git status to debug version tag

AlexanderYastrebov:
Version tag is generated based on the `git describe --tags --always --dirty`.
Starting from `9a6ffb4-dirty` it has `-dirty` suffix signalling that
working tree has local modification.

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -20,6 +20,7 @@ pipeline:
       make build.docker
   - desc: push
     cmd: |
+      git status
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         IMAGE=registry-write.opensource.zalan.do/teapot/cluster-lifecycle-manager
         VERSION=$(git describe --tags --always --dirty)


delivery.yaml
```diff
@@ -20,6 +20,7 @@ pipeline:
       make build.docker
   - desc: push
     cmd: |
+      git status
```
szuecs, line 23: why not `git diff` if VERSION has "dirty"?
```
echo $VERSION | grep -q dirty && git diff
```

AlexanderYastrebov:
```
+ git status
HEAD detached from a41f4ef
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   pkg/cluster-registry/client/clusters/list_clusters_responses.go
	modified:   pkg/cluster-registry/client/infrastructure_accounts/list_infrastructure_accounts_responses.go
	modified:   pkg/cluster-registry/client/node_pools/list_node_pools_responses.go
	modified:   pkg/cluster-registry/models/cluster.go
	modified:   pkg/cluster-registry/models/cluster_status.go
	modified:   pkg/cluster-registry/models/cluster_update.go
no changes added to commit (use "git add" and/or "git commit -a")
```

delivery.yaml
AlexanderYastrebov: git diff shows only upstaged changes and I was not sure what was changed. git status gives a better overview https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/544#issuecomment-998050897

AlexanderYastrebov:
It looks like we need to regenerate cluster registry api

PR #552, opened 2022-01-06
Drop tagging of subnets

mikkeloscar:
CLM has been tagging subnets with the ClusterID because other components (controller-manager, ingress-controller) depend on these tags in some cases for finding the right subnets for creating Load Balancers.

However, there are a couple of reasons why we don't really need/want the clusterID tag on the subnets:

1. We want to move subnet management to CloudFormation and therefore want to avoid drifting by having another component updating the tags on the subnets
2. The main reason for the ClusterID tag is to exclude subnets without those tags, however we already put the `kubernetes.io/role/elb` and `kubernetes.io/role/internal-elb` tags which are used by [CLM](https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/5c02d6941efaaecdaf7c657bf85087896569a47b/provisioner/clusterpy.go#L545-L588), [Controller-Manager](https://github.com/kubernetes/kubernetes/blob/aea7bbadd2fc0cd689de94a54e5b7b758869d691/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3431-L3522) and [Ingress-controller](https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/949a0bc169721ed9354b993f7bb810e4865bbb91/aws/ec2.go#L155-L209) for finding the right subnets to use. We can still have other non-kubernetes subnets in the accounts and just not tag them so they won't be picked up.
3. CLM was tagging ALL subnets of the VPC with the ClusterID tag so it sort of lost its meaning for 2 anyway.

Therefore we remove the logic for tagging/untagging subnets.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -210,15 +210,6 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 		return err
 	}
 
-	err = p.tagSubnets(awsAdapter, vpcID, cluster)
-	if err != nil {
-		return err
-	}
-
-	if err = ctx.Err(); err != nil {
-		return err
-	}
-
 	subnets, err := awsAdapter.GetSubnets(vpcID)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

gargravarr:
:+

gargravarr:
:+

gargravarr:
:+1:

PR #553, opened 2022-01-20
Fix error message related to amount of images found

katyanna:
If no image is found, the current error message indicates that
more than one image was found. The new if ensures a clear error
message.

Signed-off-by: Katyanna Moura <amelie.kn@gmail.com>

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -512,8 +512,11 @@ func amiID(adapter *awsAdapter, imageName, imageOwner string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to describe image with name %s and owner %s: %v", imageName, imageOwner, err)
 	}
-	if len(output.Images) != 1 {
+	if len(output.Images) > 1 {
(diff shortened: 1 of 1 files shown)


linki:
👍

provisioner/template.go
```diff
@@ -512,9 +512,12 @@ func amiID(adapter *awsAdapter, imageName, imageOwner string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to describe image with name %s and owner %s: %v", imageName, imageOwner, err)
 	}
-	if len(output.Images) != 1 {
+	if len(output.Images) > 1 {
 		return "", fmt.Errorf("more than one image found with name: %s and owner: %s", imageName, imageOwner)
 	}
+	if len(output.Images) < 1 {
```
linki, line 518: We could check for `0` here since it cannot be negative. It might be easier to understand this way.

mikkeloscar:
:+1:

katyanna:
👍

katyanna:
👍

PR #573, opened 2022-03-03
Remove Senza etcd code.

gargravarr:
Sinze we're not using Senza anymore, removes all the code used to apply the etcd Senza stack. Removes also the experimental flag to manage etcd via the Senza-less stack.  


mikkeloscar:
:+1:

PR #574, opened 2022-03-18
Drop python and senza from Docker image

mikkeloscar:
Since #573 python and senza is no longer required in the Docker image. Thus we can remove it.

This also updates the base images to latest version.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #584, opened 2022-05-05
Support any region in AWS session

mikkeloscar:
This is a small fix which enables CLM to provision clusters in any region eventhough CLM itself may be running in one particular region.

This would allow deploying a cluster in another region in an EMPTY account. In order to run clusters in two regions in the same account some changes to the etcd setup is still needed.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #585, opened 2022-05-19
feat: Add support for Karpenter managed nodes

mikkeloscar:
Changes needed for introducing Karpenter as an optional autoscaler

**This PR is just showing of the Proof of Concept as presented in the demo**

Ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5136

### DIFF
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -174,3 +174,23 @@ func (cluster *Cluster) ChannelOverrides() (map[string]string, error) {
 	}
 	return result, nil
 }
+
+func (cluster *Cluster) KarpenterPools() []*NodePool {
+	var kp []*NodePool
+	for _, n := range cluster.NodePools {
+		if n.IsKarpenter() {
+			kp = append(kp, n)
+		}
+	}
+	return kp
+}
+
+func (cluster *Cluster) ASGBackedPools() []*NodePool {
+	var cp []*NodePool
+	for _, n := range cluster.NodePools {
+		if !n.IsKarpenter() {
+			cp = append(cp, n)
+		}
+	}
+	return cp
+}
--- a/pkg/updatestrategy/aws_ec2.go
+++ b/pkg/updatestrategy/aws_ec2.go
@@ -45,9 +45,9 @@ func InstanceConfigUpToDate(instanceConfig, poolConfig *InstanceConfig) bool {
 
 type Option func(b *EC2NodePoolBackend)
 
-type NodePoolConfigGetter func(nodePool *api.NodePool) (*InstanceConfig, error)
+type NodePoolConfigGetter func(ctx context.Context, nodePool *api.NodePool) (*InstanceConfig, error)
 
-func NoopNodePoolConfigGetter(*api.NodePool) (*InstanceConfig, error) {
+func NoopNodePoolConfigGetter(context.Context, *api.NodePool) (*InstanceConfig, error) {
 	return nil, nil
 }
 
@@ -84,14 +84,14 @@ func NewEC2NodePoolBackend(clusterID string, sess *session.Session, opts ...Opti
 // The node generation is set to 'current' for nodes with up-to-date
 // userData,ImageID and tags and 'outdated' for nodes with an outdated
 // configuration.
-func (n *EC2NodePoolBackend) Get(nodePool *api.NodePool) (*NodePool, error) {
+func (n *EC2NodePoolBackend) Get(ctx context.Context, nodePool *api.NodePool) (*NodePool, error) {
 	instances, err := n.getInstances(nodePool)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list EC2 instances of the node pool: %w", err)
 	}
 
 	nodes := make([]*Node, 0)
-	nodePoolConfig, err := n.configGetter(nodePool)
+	nodePoolConfig, err := n.configGetter(ctx, nodePool)
 	if err != nil {
 		return nil, err
 	}
@@ -188,15 +188,15 @@ func (n *EC2NodePoolBackend) getInstanceConfig(i *ec2.Instance) (*InstanceConfig
 	return &config, nil
 }
 
-func (n *EC2NodePoolBackend) MarkForDecommission(*api.NodePool) error {
+func (n *EC2NodePoolBackend) MarkForDecommission(context.Context, *api.NodePool) error {
 	return nil
 }
 
-func (n *EC2NodePoolBackend) Scale(*api.NodePool, int) error {
+func (n *EC2NodePoolBackend) Scale(context.Context, *api.NodePool, int) error {
 	return nil
 }
 
-func (n *EC2NodePoolBackend) Terminate(*Node, bool) error {
+func (n *EC2NodePoolBackend) Terminate(context.Context, *Node, bool) error {
 	return nil
 }
 
--- a/pkg/updatestrategy/aws_ec2.go
+++ b/pkg/updatestrategy/aws_ec2.go
@@ -3,6 +3,7 @@ package updatestrategy
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -10,22 +11,72 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
+	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/util"
 )
 
-// EC2NodePoolBackend defines a node pool consiting of EC2 instances
+type InstanceConfig struct {
+	UserData string
+	ImageID  string
+	Tags     map[string]string
+	// TODO: Karpenter supports more parameters https://karpenter.sh/preview/concepts/node-templates/
+}
+
+// InstanceConfigUpToDate compares current and desired InstanceConfig. It compares
+// userdata, imageID and checks if the current config has all the desired tags.
+// It does NOT check if the current config has too many EC2 tags as many tags are
+// injected out of our control. This means removing a tag is not enough to
+// make the configs unequal.
+func InstanceConfigUpToDate(instanceConfig, poolConfig *InstanceConfig) bool {
+	if instanceConfig.UserData != poolConfig.UserData {
+		return false
+	}
+
+	if !util.Contains(strings.Split(poolConfig.ImageID, ","), instanceConfig.ImageID) {
+		return false
+	}
+
+	for k, v := range poolConfig.Tags {
+		if instanceValue, ok := instanceConfig.Tags[k]; !ok || v != instanceValue {
+			return false
+		}
+	}
+	return true
+}
+
+type Option func(b *EC2NodePoolBackend)
+
+type NodePoolConfigGetter func(nodePool *api.NodePool) (*InstanceConfig, error)
+
+func NoopNodePoolConfigGetter(*api.NodePool) (*InstanceConfig, error) {
+	return nil, nil
+}
+
+func WithConfigGetter(g NodePoolConfigGetter) Option {
+	return func(b *EC2NodePoolBackend) {
+		b.configGetter = g
+	}
+}
+
+// EC2NodePoolBackend defines a node pool consisting of EC2 instances
 // managed externally by some component e.g. Karpenter.
 type EC2NodePoolBackend struct {
-	ec2Client ec2iface.EC2API
-	clusterID string
+	ec2Client    ec2iface.EC2API
+	clusterID    string
+	configGetter NodePoolConfigGetter
 }
 
 // NewEC2NodePoolBackend initializes a new EC2NodePoolBackend for
 // the given clusterID and AWS session and.
-func NewEC2NodePoolBackend(clusterID string, sess *session.Session) *EC2NodePoolBackend {
-	return &EC2NodePoolBackend{
-		ec2Client: ec2.New(sess),
-		clusterID: clusterID,
+func NewEC2NodePoolBackend(clusterID string, sess *session.Session, opts ...Option) *EC2NodePoolBackend {
+	b := &EC2NodePoolBackend{
+		ec2Client:    ec2.New(sess),
+		clusterID:    clusterID,
+		configGetter: NoopNodePoolConfigGetter,
 	}
+	for _, opt := range opts {
+		opt(b)
+	}
+	return b
 }
 
 // Get gets the EC2 instances matching to the node pool by looking at node pool
@@ -40,12 +91,27 @@ func (n *EC2NodePoolBackend) Get(nodePool *api.NodePool) (*NodePool, error) {
 	}
 
 	nodes := make([]*Node, 0)
+	nodePoolConfig, err := n.configGetter(nodePool)
+	if err != nil {
+		return nil, err
+	}
 	for _, instance := range instances {
 		instanceID := aws.StringValue(instance.InstanceId)
+
+		instanceConfig, err := n.getInstanceConfig(instance)
+		if err != nil {
+			return nil, err
+		}
+		generation := currentNodeGeneration
+
+		if !InstanceConfigUpToDate(instanceConfig, nodePoolConfig) {
+			generation = outdatedNodeGeneration
+		}
+
 		node := &Node{
 			ProviderID:    fmt.Sprintf("aws:///%s/%s", aws.StringValue(instance.Placement.AvailabilityZone), instanceID),
 			FailureDomain: aws.StringValue(instance.Placement.AvailabilityZone),
-			Generation:    currentNodeGeneration,
+			Generation:    generation,
 			// not used in clc logic
 			// Ready: true,
 		}
@@ -99,44 +165,38 @@ func (n *EC2NodePoolBackend) getInstances(nodePool *api.NodePool) ([]*ec2.Instan
 	return instances, nil
 }
 
-type InstanceConfig struct {
-	UserData string
-	ImageID  string
-	Tags     map[string]string
+func (n *EC2NodePoolBackend) getInstanceConfig(i *ec2.Instance) (*InstanceConfig, error) {
+	// note: this make an extra http call to aws api for each node
+	tags := make(map[string]string, len(i.Tags))
+	for _, tag := range i.Tags {
+		tags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
+	}
+	config := InstanceConfig{
+		ImageID: aws.StringValue(i.ImageId),
+		Tags:    tags,
+	}
+	params := &ec2.DescribeInstanceAttributeInput{
+		Attribute:  aws.String("userData"),
+		DryRun:     aws.Bool(false),
+		InstanceId: i.InstanceId,
+	}
+	op, err := n.ec2Client.DescribeInstanceAttributeWithContext(context.TODO(), params)
+	if err != nil {
+		return nil, err
+	}
+	config.UserData = aws.StringValue(op.UserData.Value)
+	return &config, nil
 }
 
-// instanceConfigEqual compares current and desired InstanceConfig. It compares
-// userdata, imageID and checks if the current config has all the desired tags.
-// It does NOT check if the current config has too many EC2 tags as many tags are
-// injected out of our control. This means removing a tag is not enough to
-// make the configs unequal.
-func instanceConfigEqual(current, desired *InstanceConfig) bool {
-	if current.UserData != desired.UserData {
-		return false
-	}
-
-	if current.ImageID != desired.ImageID {
-		return false
-	}
-
-	// TODO: explain
-	for k, v := range desired.Tags {
-		if currentValue, ok := current.Tags[k]; !ok || v != currentValue {
-			return false
-		}
-	}
-	return true
-}
-
-func (n *EC2NodePoolBackend) MarkForDecommission(nodePool *api.NodePool) error {
+func (n *EC2NodePoolBackend) MarkForDecommission(*api.NodePool) error {
 	return nil
 }
 
-func (n *EC2NodePoolBackend) Scale(nodePool *api.NodePool, replicas int) error {
+func (n *EC2NodePoolBackend) Scale(*api.NodePool, int) error {
 	return nil
 }
 
-func (n *EC2NodePoolBackend) Terminate(node *Node, decrementDesired bool) error {
+func (n *EC2NodePoolBackend) Terminate(*Node, bool) error {
 	return nil
 }
 
--- a/pkg/updatestrategy/aws_ec2.go
+++ b/pkg/updatestrategy/aws_ec2.go
@@ -1,0 +1,183 @@
+package updatestrategy
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/zalando-incubator/cluster-lifecycle-manager/api"
+)
+
+// EC2NodePoolBackend defines a node pool consiting of EC2 instances
+// managed externally by some component e.g. Karpenter.
+type EC2NodePoolBackend struct {
+	ec2Client ec2iface.EC2API
+	clusterID string
+}
+
+// NewEC2NodePoolBackend initializes a new EC2NodePoolBackend for
+// the given clusterID and AWS session and.
+func NewEC2NodePoolBackend(clusterID string, sess *session.Session) *EC2NodePoolBackend {
+	return &EC2NodePoolBackend{
+		ec2Client: ec2.New(sess),
+		clusterID: clusterID,
+	}
+}
+
+// Get gets the EC2 instances matching to the node pool by looking at node pool
+// tag.
+// The node generation is set to 'current' for nodes with up-to-date
+// userData,ImageID and tags and 'outdated' for nodes with an outdated
+// configuration.
+func (n *EC2NodePoolBackend) Get(nodePool *api.NodePool) (*NodePool, error) {
+	instances, err := n.getInstances(nodePool)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list EC2 instances of the node pool: %w", err)
+	}
+
+	nodes := make([]*Node, 0)
+	for _, instance := range instances {
+		instanceID := aws.StringValue(instance.InstanceId)
+		node := &Node{
+			ProviderID:    fmt.Sprintf("aws:///%s/%s", aws.StringValue(instance.Placement.AvailabilityZone), instanceID),
+			FailureDomain: aws.StringValue(instance.Placement.AvailabilityZone),
+			Generation:    currentNodeGeneration,
+			// not used in clc logic
+			// Ready: true,
+		}
+
+		nodes = append(nodes, node)
+	}
+
+	// We only set Generation and Nodes as nothing else is needed by the
+	// CLC strategy
+	return &NodePool{
+		Generation: currentNodeGeneration,
+		Nodes:      nodes,
+	}, nil
+}
+
+// getInstances lists all running instances of the node pool.
+func (n *EC2NodePoolBackend) getInstances(nodePool *api.NodePool) ([]*ec2.Instance, error) {
+	params := &ec2.DescribeInstancesInput{
+		Filters: []*ec2.Filter{
+			{
+				Name: aws.String("tag:" + clusterIDTagPrefix + n.clusterID),
+				Values: []*string{
+					aws.String(resourceLifecycleOwned),
+				},
+			},
+			{
+				Name: aws.String("tag:" + nodePoolTag),
+				Values: []*string{
+					aws.String(nodePool.Name),
+				},
+			},
+		},
+	}
+
+	instances := make([]*ec2.Instance, 0)
+	err := n.ec2Client.DescribeInstancesPagesWithContext(context.TODO(), params, func(output *ec2.DescribeInstancesOutput, lastPage bool) bool {
+		for _, reservation := range output.Reservations {
+			for _, instance := range reservation.Instances {
+				switch aws.StringValue(instance.State.Name) {
+				case ec2.InstanceStateNameRunning, ec2.InstanceStateNamePending, ec2.InstanceStateNameStopped:
+					instances = append(instances, instance)
+				}
+			}
+		}
+		return true
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return instances, nil
+}
+
+type InstanceConfig struct {
+	UserData string
+	ImageID  string
+	Tags     map[string]string
+}
+
+// instanceConfigEqual compares current and desired InstanceConfig. It compares
+// userdata, imageID and checks if the current config has all the desired tags.
+// It does NOT check if the current config has too many EC2 tags as many tags are
+// injected out of our control. This means removing a tag is not enough to
+// make the configs unequal.
+func instanceConfigEqual(current, desired *InstanceConfig) bool {
+	if current.UserData != desired.UserData {
+		return false
+	}
+
+	if current.ImageID != desired.ImageID {
+		return false
+	}
+
+	// TODO: explain
+	for k, v := range desired.Tags {
+		if currentValue, ok := current.Tags[k]; !ok || v != currentValue {
+			return false
+		}
+	}
+	return true
+}
+
+func (n *EC2NodePoolBackend) MarkForDecommission(nodePool *api.NodePool) error {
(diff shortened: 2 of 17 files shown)


pkg/updatestrategy/aws_ec2.go
```diff
@@ -0,0 +1,245 @@
...
+			ImageID:  aws.StringValue(instance.ImageId),
+			Tags:     make(map[string]string, len(instance.Tags)),
+		}
+
+		for _, tag := range instance.Tags {
+			currentInstanceConfig.Tags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
+		}
+
+		ltParams := &ec2.DescribeLaunchTemplateVersionsInput{
+			LaunchTemplateName: aws.String(awsValidID(n.clusterID) + "-" + nodePool.Name),
```
mikkeloscar, line 77: Does this still work/match after updating the kubernetes-on-aws PR to use the `AWSNodeTemplate` resource for creating the launch template?

provisioner/node_pools.go
```diff
@@ -123,13 +188,124 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
...
+	_, err = p.kubectlExecute(ctx, []string{"apply"}, template)
+	if err != nil {
+		return errors.Wrapf(err, "kubectl apply failed for node pool %s", nodePool.Name)
+	}
+	return nil
+}
+
+func (p *KarpenterNodePoolProvisioner) Reconcile(ctx context.Context, updater updatestrategy.UpdateStrategy) error {
+	karpenterPools := p.cluster.KarpenterPools()
+	output, err := p.kubectlExecute(ctx, []string{"get", "provisioner", "--no-headers", "-o", "custom-columns=':metadata.name'"}, "")
```
mikkeloscar, line 243: IMO Let's use a proper client-go for interacting with the API! We only use `kubectl` binary for the `apply` case because it depend on client-side logic in `kubectl`. For getting/deleting provisioners we don't need it.

provisioner/template_test.go
```diff
@@ -772,6 +772,18 @@ func TestZoneDistributedNodePoolGroupsDedicated(t *testing.T) {
...
+		{
+			Name:        "valid-1",
+			Profile:     "worker-karpenter",
+			ConfigItems: map[string]string{"labels": "dedicated=valid-karpenter", "taints": "dedicated=valid-karpenter:NoSchedule"},
+		},
+		{
+			Name:        "valid-2",
+			Profile:     "worker-karpenter",
+			ConfigItems: map[string]string{"labels": "dedicated=valid-karpenter", "taints": "dedicated=valid-karpenter:NoSchedule"},
+		},
```
linki, line 785: I think we don't need two test pools here because we won't have this combination in production. The two pools just above are for the fallback functionality (two pools, same dedicated label) which karpenter already provides within just one pool.

provisioner/template.go
```diff
@@ -655,6 +664,7 @@ func poolsDistributed(dedicated string, pools []*api.NodePool) bool {
 //   - are correctly configured with regards to the labels and taints
 //   - don't have AZ restrictions
 //   - use the worker-splitaz profile.
+//   - use the worker-karpenter profile.
```
linki, line 667: Based on the implementation should this rather be "- use the worker-splitaz or worker-karpenter profile."?

provisioner/template.go
```diff
@@ -135,6 +136,12 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"zoneDistributedNodePoolGroups": zoneDistributedNodePoolGroups,
 		"certificateExpiry":             certificateExpiry,
 		"sumQuantities":                 sumQuantities,
+		"awsValidID":                    awsValidID,
+		"karpenterNodePools":            karpenterNodePools,
+		"capacityTypes": func() []string {
+			return []string{"spot", "on-demand"}
```
linki, line 142: Since it's just a static value, can we avoid the function and put that into the templates?

provisioner/clusterpy.go
```diff
@@ -625,6 +630,21 @@ func (p *clusterpyProvisioner) Decommission(ctx context.Context, logger *log.Ent
...
 
+	// decommission each node pool
+	// TODO: for now only karpenter pools are supported. Other pools are
+	// decommissioned by the stack deletion logic below.
+	// Would be cleaner to have a per node pool decommissioning logic.
+	for _, nodePool := range cluster.NodePools {
+		if nodePool.Profile == karpenterNodePoolProfile {
+			logger.Infof("Decommissioning Node Pool: %s (profile: %s)", nodePool.Name, nodePool.Profile)
+			nodePoolBackend := updatestrategy.NewEC2NodePoolBackend(cluster.ID, awsAdapter.session)
+			err := nodePoolBackend.Decommission(ctx, nodePool)
```
linki, line 641: According to the docs the recommended approach to decommission karpenter nodes is to delete the `Node` object via the Kubernetes API. Worked fine when I tried it.

api/cluster.go
```diff
@@ -174,3 +174,23 @@ func (cluster *Cluster) ChannelOverrides() (map[string]string, error) {
...
+		}
+	}
+	return kp
+}
+
+func (cluster *Cluster) ClusterAutoscalerPools() []*NodePool {
+	var cp []*NodePool
+	for _, n := range cluster.NodePools {
+		if !n.IsKarpenter() {
+			cp = append(cp, n)
```
linki, line 192: Do we care that this would include master node pools which are not managed by CA? We could also rename `ClusterAutoscalerPools()` to something like `ASGBackedPools` which would be factually correct for masters and legacy workers.

provisioner/clusterpy.go
myaser: Actually, deleting the provisioner resource should be enough, and karpenter shall delete the nodes automatically.
I see your point that terminating the nodes from the AWS API is not needed.

but what do you think
1. only delete the provisioner resource and leave it to karpenter to clean up the nodes
2. delete the provisioner resource and delete the node objects from kubectl
3. delete the provisioner resource and label the nodes for cluster-lifecycle-manager to decommission

provisioner/template_test.go
myaser: in production, it may happen to have two pools with fallback (provisioner `spec.weight` feature)

For example if we want to fallback *only* in case of unavailable capacity of the higher priority pool

provisioner/clusterpy.go
linki: I prefer 1. over 2.

Regarding 3: We definitely need to check compatibility with cluster-lifecycle-manager (CLC). We have some logic there (or in CA) that's specific to Zalando (e.g. the forced eviction of Pods after 3 days). Ideally karpenter's way of cleaning up (which respects PDBs already) should be the only thing we need (maybe we need to contribute back if there's something missing). 

Otherwise we need to somehow make it work with CLC.

myaser: I will go with 1 for now, until we do better discovery

myaser: actually, there are 2 different scenarios where we need to decommission a node pool

-a) configuring the cluster with `zregistry` deleting a node pool
-b) decommissioning the whole cluster

what we discussed fits more for the first scenario, for the later we need to actively delete the nodes in order not to wait until karpenter controller cleans them

linki: We can also leave it like it is for now if it works.

pkg/updatestrategy/aws_ec2.go
myaser: No, it does not work; you are right.
I think I can remove this part right?

provisioner/node_pools.go
```diff
@@ -123,13 +201,164 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
...
+	}
+	_, err = p.kubectlExecute(ctx, []string{"apply"}, template)
+	if err != nil {
+		return errors.Wrapf(err, "kubectl apply failed for node pool %s", nodePool.Name)
+	}
+	return nil
+}
+
+func (p *KarpenterNodePoolProvisioner) isKarpenterInstalled(ctx context.Context) (bool, error) {
+	_, dynamicClient, mapper, err := kubernetes.InitClients(p.cluster.APIServerURL, p.tokenSource)
```
linki, line 257: Should we use an existing client here instead of initializing a new one each time we call this function?

provisioner/node_pools.go
```diff
@@ -123,13 +201,164 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
...
+
+	_, dynamicClient, mapper, err := kubernetes.InitClients(p.cluster.APIServerURL, p.tokenSource)
+	if err != nil {
+		return err
+	}
+	provisionerGvr, err := kubernetes.ResolveKind(mapper, "provisioner.karpenter.sh")
+	if err != nil {
+		return err
+	}
+	nodeTemplateGvr, err := kubernetes.ResolveKind(mapper, "awsnodetemplates.karpenter.k8s.aws")
```
linki, line 297: Let's put these values into `const`s.

provisioner/node_pools.go
```diff
@@ -123,13 +201,164 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
...
+	existingProvisioners, err := dynamicClient.Resource(provisionerGvr).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, p := range existingProvisioners.Items {
+		if !inNodePoolList(&api.NodePool{Name: p.GetName()}, karpenterPools) {
+			err := dynamicClient.Resource(provisionerGvr).Delete(ctx, p.GetName(), metav1.DeleteOptions{})
+			if err != nil {
+				return err
```
linki, line 310: If we fail here, we could end up just deleting the provisioner (and not the awstemplate). Do we care about that?

provisioner/clusterpy.go
```diff
@@ -328,70 +324,74 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
...
-			return err
-		}
+	if err = ctx.Err(); err != nil {
+		return err
+	}
 
-		if err = ctx.Err(); err != nil {
-			return err
-		}
+	err = p.apply(ctx, logger, cluster, deletions, manifests)
```
linki, line 370: This changes the update sequece from `master->workers->manifests` to `master->manifests->workers`.

Is there a way to keep the old flow?

myaser: `karpenter worker pools (provisioners)` depends on `karpenter CRDs & karpenter validation webhook a container in the karpenter controller).`
I tried this: `master -> apply necessary karpenter manifests only -> workers -> manifests`, this did not work as karpenter controller has dependency on other manifests. so I had to bring all the manifests before the worker nodes

as an alternative, we can experiment creating the karpenter provisioners workers before their dependencies (injecting them to etcd cluster directly to bypass api-server validation) or other ways.

I thought we discussed this and agreed upon this approach of `master->manifests->workers`

pkg/updatestrategy/aws_ec2.go
mikkeloscar: Having this removed basically means that nodes are always considered up to date and would not be rolled by CLC strategy.

Is this handled another way?

provisioner/clusterpy.go
```diff
@@ -184,7 +180,7 @@ func (p *clusterpyProvisioner) propagateConfigItemsToNodePools(cluster *api.Clus
 // Provision provisions/updates a cluster on AWS. Provision is an idempotent
 // operation for the same input.
 func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry, cluster *api.Cluster, channelConfig channel.Config) error {
-	awsAdapter, updater, nodePoolManager, err := p.prepareProvision(logger, cluster, channelConfig)
+	awsAdapter, updater, _, err := p.prepareProvision(logger, cluster, channelConfig)
```
mikkeloscar, line 183: If the `nodePoolManager` is not used from `prepareProvision`, then it can be removed from the function.

provisioner/clusterpy.go
mikkeloscar: Being this close to Cyber Week I would be worried about introducing this change now as it would apply to ALL clusters also where we don't test karpenter. Can we somehow do it via a config item so this order is only the case when karpenter support is enabled. Allowing us to move forward with testing karpenter in test, but stay without changes to behavior in production.

provisioner/node_pools.go
```diff
@@ -123,13 +223,150 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
...
+	}
+	existingProvisioners, err := p.kubeClient.Resource(provisionerGvr).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, pr := range existingProvisioners.Items {
+		if !inNodePoolList(&api.NodePool{Name: pr.GetName()}, karpenterPools) {
+			_ = p.kubeClient.Resource(provisionerGvr).Delete(ctx, pr.GetName(), metav1.DeleteOptions{})
+			_ = p.kubeClient.Resource(nodeTemplateGvr).Delete(ctx, pr.GetName(), metav1.DeleteOptions{})
```
linki, line 323: Ignoring the error doesn't feel good either. Let's use the old approach or at least print the error.

I was just worried about the case where the `provisioner` gets deleted but the `awstemplate` fails. It will never retry in this case and it will never be cleaned up. Not a big deal I guess.

provisioner/node_pools.go
```diff
@@ -123,13 +223,150 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
 	return nil
 }
 
-func supportsT2Unlimited(instanceTypes []string) bool {
-	for _, instanceType := range instanceTypes {
-		if !strings.HasPrefix(instanceType, "t2.") {
-			return false
+func (p *KarpenterNodePoolProvisioner) kubectlExecute(ctx context.Context, args []string, stdin string) (string, error) {
```
mikkeloscar, line 226: nit: we duplicate this logic from `clusterpy.go`. Would be good to turn it into a generic function. Can be done later.

provisioner/node_pools.go
```diff
@@ -123,13 +223,150 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
...
+	return nil
+}
+
+func (p *KarpenterNodePoolProvisioner) isKarpenterInstalled(ctx context.Context) (bool, error) {
+	gvr, err := kubernetes.ResolveKind(p.kubeMapper, crd)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = p.kubeClient.Resource(gvr).Get(ctx, karpenterProvisionerResource, metav1.GetOptions{})
```
mikkeloscar, line 284: nit: Since this is moved to the `kubernetes` package it could also be an opportunity to make the interface a bit nicer. Instead of an interface that returns multiple clients via `InitClients` it could have a client that hides the internal details of `ResolveKind` etc.

So instead of calling first `ResolveKind` and then `kubeClient.Resource(gvr).Get(ctx, karpenterProvisionerResource, metav1.GetOptions{})` you could have a method `GetCRD(ctx, resourcename, metav1.GetOptions{})` which does the `ResolveKind` internally.

This can ofc. be done later. Just came to light because of the many Exported functions in `kubernetes.go` which is hinting at a better abstraction being possible.

provisioner/node_pools.go
mikkeloscar, line 323: ignoring errors seem wrong :) Something didn't work so we should at least log it or maybe fail here.

provisioner/node_pools.go
```diff
@@ -219,62 +456,13 @@ func (p *AWSNodePoolProvisioner) Reconcile(ctx context.Context, updater updatest
...
-	writer := gzip.NewWriter(&gzipBuffer)
-	defer writer.Close()
-	_, err = writer.Write([]byte(rendered))
-	if err != nil {
-		return "", err
-	}
-	err = writer.Close()
-	if err != nil {
-		return "", err
+func supportsT2Unlimited(instanceTypes []string) bool {
```
mikkeloscar, line 459: Can you add a comment about this? I vaguely remember a comment about it, but looking at it now I have no clue why we would need this function? :)

provisioner/clusterpy.go
myaser: Alternatively, as part of the cluster decommission, we can start by deleting the karpenter provisioners and wait until karpenter controller delete all nodes.

then we downscale kube-system controllers
this is a bit different behavior than what we have 2day

pkg/updatestrategy/aws_ec2.go
```diff
@@ -0,0 +1,183 @@
...
+	ImageID  string
+	Tags     map[string]string
+}
+
+// instanceConfigEqual compares current and desired InstanceConfig. It compares
+// userdata, imageID and checks if the current config has all the desired tags.
+// It does NOT check if the current config has too many EC2 tags as many tags are
+// injected out of our control. This means removing a tag is not enough to
+// make the configs unequal.
+func instanceConfigEqual(current, desired *InstanceConfig) bool {
```
mikkeloscar, line 113: This function is not called. I think it should be called as part of the `Get` method to filter new/old nodes?

pkg/updatestrategy/aws_ec2.go
```diff
@@ -0,0 +1,231 @@
+package updatestrategy
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/zalando-incubator/cluster-lifecycle-manager/provisioner"
```
mikkeloscar, line 8: This looks like a circular dependency. I imagine that is why the build is failing?

myaser:
I tested the recent changes regarding CLC integration on a pet cluster, and I see it working
```
INFO[0047] Initializing update of node pool 'karpenter-default'  strategy=clc
E0405 13:55:01.566436   59315 memcache.go:196] couldn't get resource list for custom.metrics.k8s.io/v1beta1: Got empty response for: custom.metrics.k8s.io/v1beta1
E0405 13:55:01.566666   59315 memcache.go:196] couldn't get resource list for external.metrics.k8s.io/v1beta1: Got empty response for: external.metrics.k8s.io/v1beta1
INFO[0300] Waiting for decommissioning of old nodes (1 left)  node-pool=karpenter-default strategy=clc
INFO[0318] Waiting for decommissioning of old nodes (1 left)  node-pool=karpenter-default strategy=clc
INFO[0402] Waiting for decommissioning of old nodes (1 left)  node-pool=karpenter-default strategy=clc
INFO[0422] Node pool 'karpenter-default' successfully updated  strategy=clc
```

note:
I had to workaround [this Karpenter issue](https://github.com/aws/karpenter/issues/3697) by manually editing the tag to match the clusterName. This tag is used by CLC to select the nodes to be updated

the issue is [being solved](https://github.com/aws/karpenter/pull/3701) and I will update to the latest version when released. but this will be in the k8s-on-aws repo. no changes for the CLM

provisioner/clusterpy.go
```diff
@@ -333,38 +334,73 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
...
 	}
 
-	if err = ctx.Err(); err != nil {
-		return err
+	//if err = ctx.Err(); err != nil {
+	//	return err
+	//}
+
+	if karpenterProvisioner.isKarpenterEnabled() {
+		err = p.apply(ctx, logger, cluster, deletions, manifests)
```
mikkeloscar, line 369: @linki @myaser I remember we discussed this back in 2022 and we thought about looking at previous kubernetes-on-aws changes where the order (node rotation, apply manifests) mattered, to see if changing this order would be a problem.

Did we do this?

provisioner/clusterpy.go
```diff
@@ -333,38 +334,73 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
...
 	err = nodePoolGroups["masters"].provisionNodePoolGroup(ctx, values, updater, cluster, p.applyOnly)
 	if err != nil {
 		return err
 	}
 
-	if err = ctx.Err(); err != nil {
-		return err
+	//if err = ctx.Err(); err != nil {
+	//	return err
+	//}
```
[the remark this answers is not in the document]
myaser: I moved them into `func (npg *nodePoolGroup) provisionNodePoolGroup`

so I thought they are redundant. kept them commented during review to start discussion as I am not sure

what do you think?

provisioner/clusterpy.go
myaser: Not really
We would need a list of typical updates and then check if an order change could cause a problem
I don't think we have such a list

do we audit log zregistry commands? would be the place where we build this list, otherwise it will be only guessing

mikkeloscar: For these more complex updates we often the naming pattern of **Change X [1/2]** or similar:

E.g. like these:

image

My proposal would be to look at such PRs and understand if they would work or if the order could be constructed differently such that they would. It want find _all_ cases but can be a good check to think about the potential consequences.

(It seems not easy to search for this pattern in github search though 😥)

myaser: using github CLI to list all PRs
```bash
gh pr list --state all --limit 5777
```
and filtering the output using regex, I compiled this list of 61 occurance
```
5663	Enable 2k nodes by default stage [2/2]	enable-2k-nodes-stage2	MERGED	2023-01-09 17:02:37 +0000 UTC
5643	Enable 2k nodes by default stage1 [1/2]	enable-2k-nodes-stage1	MERGED	2023-01-03 11:12:47 +0000 UTC
5150	Consolidate e2e application [1/2]	cleanup-consolidation-1	MERGED	2022-05-30 16:27:59 +0000 UTC
4990	Consolidate DaemonSet applications/components [3/3]	consolidate-ds-apps-3	MERGED	2022-02-14 15:03:52 +0000 UTC
4975	Consolidate DaemonSet applications/components [2/3]	consolidate-ds-apps-2	MERGED	2022-02-08 12:43:24 +0000 UTC
4945	Consolidate prometheus into kubernetes application [3/3]	prometheus-consolidation-3	MERGED	2022-01-28 09:33:55 +0000 UTC
4939	Set custom cluster tag on egress stacks [2/2]	egress-controller-cluster-id-tag	MERGED	2022-01-26 14:43:29 +0000 UTC
4934	Consolidate prometheus into kubernetes application [2/3]	prometheus-consolidation-2	MERGED	2022-01-25 16:05:30 +0000 UTC
4910	Set custom cluster tag on egress stacks [1/2]	egress-cluster-tag	MERGED	2022-01-19 14:33:09 +0000 UTC
4900	Consolidate prometheus into kubernetes application [1/3]	prometheus-consolidation	MERGED	2022-01-14 17:38:47 +0000 UTC
4898	Consolidate DaemonSet applications/components [1/3]	consolidate-ds-apps	MERGED	2022-01-14 16:43:38 +0000 UTC
4704	Skipper/routesrv2 [2/2]	skipper/routesrv2	MERGED	2021-10-06 16:33:14 +0000 UTC
4702	Skipper/routesrv [1/2]	skipper/routesrv	MERGED	2021-10-05 18:10:07 +0000 UTC
4393	Remove spotio node pool from e2e [1/2]	rm-spotio-0	MERGED	2021-06-23 11:33:16 +0000 UTC
4392	Remove spotio support [2/2]	rm-spotio	MERGED	2021-06-23 10:10:07 +0000 UTC
4327	Drop worker_shared_secret from the e2e script [2/2]	drop-shared-secret-2	MERGED	2021-05-31 15:51:52 +0000 UTC
4326	Drop node_auth and worker_shared_secret from the manifests [1/2]	drop-shared-secret	MERGED	2021-05-31 15:50:21 +0000 UTC
4284	credentials-provider: fully migrate to restricted tokens [2/2]	credprov-no-old-tokens	MERGED	2021-05-05 12:48:05 +0000 UTC
4260	node_auth: enabled->exclusive [2/2]	node-auth-exclusive	MERGED	2021-04-28 10:49:23 +0000 UTC
4259	node_auth: supported->enabled [1/2]	node-auth-enabled	MERGED	2021-04-28 10:47:19 +0000 UTC
4243	[2/3] Reduce skipper metrics dimension	skipper-new-counter-metric	MERGED	2021-04-19 17:20:39 +0000 UTC
4235	[1/3] Update skipper and use new counter metric	skipper-new-counter-metric	MERGED	2021-04-19 05:47:48 +0000 UTC
4201	Master SG: drop the ingress rules that were only used for metrics [2/2]	close-unnecessary-ports	MERGED	2021-03-29 16:11:57 +0000 UTC
4199	Cleanup the metrics for the apiserver components [1/2]	apiserver-container-metrics	MERGED	2021-03-29 15:59:50 +0000 UTC
4112	E2E: generate a per-run WORKER_SHARED_SECRET [3/3]	e2e-auth-cleanup	MERGED	2021-03-04 15:10:12 +0000 UTC
4109	E2E: support OAuth tokens as well [1/3]	e2e-auth-pre	MERGED	2021-03-04 11:59:17 +0000 UTC
4108	E2E: use an OAuth token instead of a static secret [2/3]	e2e-auth	MERGED	2021-03-04 11:51:00 +0000 UTC
4092	Support scraping Prometheus metrics via pod entities [2/2]	prometheus-pod-entity-enable	MERGED	2021-02-25 15:02:07 +0000 UTC
4086	Support scraping Prometheus metrics via pod entities [1/2]	prometheus-pod-entities	MERGED	2021-02-25 08:16:52 +0000 UTC
4013	Switch back to legacy OIDC path for migration [2/2]	add-legacy-oidc-keys-path-part-2	MERGED	2021-02-03 15:48:06 +0000 UTC
4011	Still support old OIDC keys path [1/2]	add-legacy-oidc-keys-path	MERGED	2021-02-03 14:43:10 +0000 UTC
3947	Add kube-node-ready-controller [2/2] (test)	enable-kube-node-ready-controller-test	MERGED	2021-01-19 09:34:06 +0000 UTC
3907	Add kube-node-ready-controller [1/2]	kube-node-ready-controller	MERGED	2021-01-08 16:49:01 +0000 UTC
3886	Improve the e2e steps [2/2]	better-certs-2	MERGED	2021-01-06 15:56:46 +0000 UTC
3885	Improve the e2e steps [1/2]	better-certs	MERGED	2021-01-06 12:43:39 +0000 UTC
3871	Drop admission controller proxy [2/2]	drop-proxy-2	MERGED	2021-01-04 10:46:52 +0000 UTC
3870	Drop admission controller proxy [1/2]	drop-proxy-1	MERGED	2021-01-04 10:31:51 +0000 UTC
3700	Support permanent scale-up for CA v1.18 [1/2]	cluster_autoscaler_1_18_2	CLOSED	2020-10-30 09:56:45 +0000 UTC
3396	Remove all traces of API server ELBv1 (part 2/2)	remove-elb-entirely	MERGED	2020-07-07 15:13:44 +0000 UTC
3393	Remove all traces of API server ELBv1 (part 1.9/2)	remove-elb-entirely	MERGED	2020-07-07 12:35:34 +0000 UTC
3373	Remove all traces of API server ELBv1 (part 1/2)	remove-elbv1	MERGED	2020-07-01 10:08:32 +0000 UTC
2891	[1/2] CDP controller role: allow the use of v2 tokens	cdp-controller-v2	MERGED	2020-01-23 12:16:29 +0000 UTC
2865	RBAC: rollout [2/2]	revert-2864-revert-2862-revert-2861-revert-2456-webhook-deny-only	MERGED	2020-01-14 17:51:23 +0000 UTC
2864	Revert "Revert "Revert "RBAC: rollout [2/2]"""	revert-2862-revert-2861-revert-2456-webhook-deny-only	MERGED	2020-01-14 15:22:25 +0000 UTC
2862	Revert "Revert "RBAC: rollout [2/2]""	revert-2861-revert-2456-webhook-deny-only	MERGED	2020-01-14 12:59:01 +0000 UTC
2861	Revert "RBAC: rollout [2/2]"	revert-2456-webhook-deny-only	MERGED	2020-01-14 11:14:16 +0000 UTC
2844	Update audittrail-adapter to work with v1 audit events instead of v1beta1 [1/2]	audittrail-adapter-new-api	MERGED	2020-01-08 13:14:09 +0000 UTC
2749	Limit nodes considered for ingress load balancer based on taints [1/2]	ingress-controller-targets	MERGED	2019-12-04 16:11:29 +0000 UTC
2520	Revert "logging-agent: support new node configuration [1/2]"	revert-logging-change	MERGED	2019-09-16 13:14:07 +0000 UTC
2515	logging-agent: support new node configuration [1/2]	migrate-logging-agent	MERGED	2019-09-13 16:02:45 +0000 UTC
2510	RBAC: rollout [1/2]	update-rbac-roles	MERGED	2019-09-12 11:11:18 +0000 UTC
2456	RBAC: rollout [2/2]	webhook-deny-only	MERGED	2019-08-27 15:41:35 +0000 UTC
2202	RBAC zalando iam 2 [2/2]	rbac-zalando-iam-2	MERGED	2019-06-06 09:57:18 +0000 UTC
2200	Add RBAC for Zalando IAM users [1/2]	rbac-zalando-iam	MERGED	2019-06-05 13:58:41 +0000 UTC
2170	Switch kubelets to authenticated mode [3/3]	kubelet-auth-part-3	MERGED	2019-05-27 09:34:57 +0000 UTC
2167	Kubelet: use authenticated API by default [2/3]	kubelet-auth-part-2	MERGED	2019-05-24 08:25:57 +0000 UTC
2153	Kubelet: switch to config instead of arguments, use the new certificate [1/3]	kubelet-auth-part-1	MERGED	2019-05-22 10:17:56 +0000 UTC
1183	dnsmasq-node to 1.9 (production 1/2)	dnsmasq-to-1.9	MERGED	2018-06-19 12:02:27 +0000 UTC
1181	Flannel to 1.9 (production 2/2)	flannel-to-1.9	MERGED	2018-06-19 10:20:56 +0000 UTC
1121	Remove obsolete stack parameters (Legacy pools step 4/4)	remove-static-pools2	MERGED	2018-06-05 10:06:39 +0000 UTC
1091	Remove obsolete static node pools (Legacy pools step 2/4)	remove-static-pools	MERGED	2018-05-30 13:10:11 +0000 UTC
```

provisioner/node_pools.go
```diff
@@ -343,3 +456,43 @@ func nodePoolStackToNodePool(stack *cloudformation.Stack) *api.NodePool {
 	}
 	return nodePool
 }
+
+func KarpenterNodePoolConfigGetter(kubeClient *kubernetes.ClientsCollection) updatestrategy.NodePoolConfigGetter {
```
mikkeloscar, line 460: It would be good practice to pass a `ctx` here and then use that in the client calls below where it currently uses `context.TODO()`.

mikkeloscar: _correction_: It would be the `updatestrategy.NodePoolConfigGetter` that should accept a `context` as the first argument.

provisioner/node_pools.go
```diff
@@ -123,13 +223,150 @@ func (p *AWSNodePoolProvisioner) Provision(ctx context.Context, nodePools []*api
...
+func (p *KarpenterNodePoolProvisioner) Reconcile(ctx context.Context, updater updatestrategy.UpdateStrategy) error {
+	karpenterPools := p.cluster.KarpenterPools()
+	installed, err := p.isKarpenterInstalled(ctx)
+	if err != nil {
+		return err
+	}
+	if !installed {
+		// skip if karpenter is not installed
+		p.logger.Infof("skipping reconcilation of karpenter node pools as karpenter is not installed")
+		return nil
```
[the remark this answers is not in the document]
mikkeloscar: I unresolved this because I still see this as questionable

provisioner/node_pools.go
mikkeloscar: What is the effect of having a leaked awstemplate resource?

provisioner/node_pools.go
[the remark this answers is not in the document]
myaser: isKarpenterEnabled checks if the feature flag is set to true.
I thought it would be simpler to skip the operation if the flag is false.

I will move the check into `clusterpy.go` only call reconcile if karpenter feature flag is enabled.

provisioner/node_pools.go
myaser: The most important resource is the provisioner. this is a synonym to node-pool in our setup. if the provisioner is not there the controller will delete all the nodes belonging to it and will stop creating more nodes

awstemplate is an abstraction to AWS launch template. it must be referenced in a provisioner to be used. so the impact is some bytes of storage in ETCD.

in addition, the CLM will fail and retry,  we can then see the problem and address the problematic resource to unblock CLM

provisioner/node_pools.go
myaser: will do

myaser: done

provisioner/node_pools.go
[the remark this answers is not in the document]
myaser: done

provisioner/node_pools.go
myaser: its low effort, so I did it.

mikkeloscar:
:+1:

myaser:
👍🏻

linki:
👏

PR #590, opened 2022-06-22
Expose architecture of node pool to the templates

linki:
~~This adds a template function called `amiIDWithArch` that works similar to `amdID` in that it finds an AMI ID (`ami-364sgy3` etc.) from a nicer name and an account ID.~~

Since different CPU architectures require different AMIs and different AMIs need to be called differently ~~, this function was added.~~

~~It takes an additional parameter that contains the architecture of the node pool and replaces any occurrence of `$ARCH` in the AMI's name with that value. If no such variable is defined then this works similar to `amiID`.~~

For intended usage, see this PR: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5200

In a nutshell: You set the AMI similar to how we currently do it:

```yaml
kuberuntu_image_v1_21_amd64: {{ amiID zalando-ubuntu-kubernetes-production-v1.21.12-amd64-master-229 "<owner>" }}
kuberuntu_image_v1_21_arm64: {{ amiID zalando-ubuntu-kubernetes-production-v1.21.12-arm64-master-229 "<owner>" }}
```

~~The only difference is that it's not within `{{ }}` so the default is the string an not the already resolved AMI ID. This also makes it easier to overwrite the AMI in a node pool's config item because you can (must) use the name.~~

~~Secondly it contains an _optional_ `$ARCH` variable that's substituted with the node pool's architecture and works well with our new naming scheme at https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu. Not specifying `$ARCH` will leave the string as is which should make the migration easy.~~

The only addition of this PR now is that it adds an `.Architecture` field to the `InstanceType` that is available in the templates. It can be used to pick the right config item (shown above) for the node pool.

### DIFF
--- a/pkg/aws/instance_info_test.go
+++ b/pkg/aws/instance_info_test.go
@@ -26,7 +26,8 @@ func (mock *mockEC2) DescribeInstanceTypesPages(_ *ec2.DescribeInstanceTypesInpu
 				},
 				InstanceStorageInfo: &ec2.InstanceStorageInfo{},
 				ProcessorInfo: &ec2.ProcessorInfo{
-					SupportedArchitectures: []*string{aws.String("x86_64")},
+					// Test that unsupported architectures are correctly ignored.
+					SupportedArchitectures: []*string{aws.String("x86_64_mac"), aws.String("i386"), aws.String("x86_64")},
 				},
 			},
 			{
@@ -405,6 +406,26 @@ func TestSyntheticInstanceInfo(t *testing.T) {
 			instanceTypes: []string{"m6g.xlarge"},
 			expectedInstance: Instance{
 				InstanceType: "m6g.xlarge",
+				VCPU:         4,
+				Memory:       17179869184,
+				Architecture: "arm64",
+			},
+		},
+		{
+			name:          "multiple types, different architectures, amd64 architecture first",
+			instanceTypes: []string{"m5.xlarge", "m6g.xlarge"},
+			expectedInstance: Instance{
+				InstanceType: "<multiple>",
+				VCPU:         4,
+				Memory:       17179869184,
+				Architecture: "amd64",
+			},
+		},
+		{
+			name:          "multiple types, different architectures, arm64 architecture first",
+			instanceTypes: []string{"m6g.xlarge", "m5.xlarge"},
+			expectedInstance: Instance{
+				InstanceType: "<multiple>",
 				VCPU:         4,
 				Memory:       17179869184,
 				Architecture: "arm64",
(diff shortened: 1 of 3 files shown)


myaser:
LGTM
but maybe we can use a map or a select statement to resolve the arm64 & amd64 AMIs.
it makes both AMIs defined explicitly in the YAML file and takes the guesswork out

pkg/aws/instance_info_test.go
```diff
@@ -337,6 +397,17 @@ func TestSyntheticInstanceInfo(t *testing.T) {
...
+			},
+		},
+		{
+			name:          "one type with arm64",
+			instanceTypes: []string{"m6g.xlarge"},
+			expectedInstance: Instance{
+				InstanceType: "m6g.xlarge",
+				VCPU:         4,
+				Memory:       17179869184,
+				Architecture: "arm64",
```
mikkeloscar, line 410: What about if we add multiple instances types of different arch?

linki: Please don't 😆

mikkeloscar: I understand we can't/won't support that, but will it error out or will it just go to the first type found? I wonder how hard it would be to debug if done by accident basically.

mikkeloscar: Ok, I guess that is good enough for now, the UX validation/improvements should be done anyway closer to the operators (us). I just wanted to make sure it doesn't fail only when AWS tries to create an instance from an ASG or so.

linki: It leaves the question what happens when the AMI and first instance type are fine but the second instance type in the pool (e.g. spot for us) doesn't work with the AMI. Will need to check if AWS already blocks this for us.

linki:
@myaser Thanks for the suggestion! I ended with this: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5207/files

Looks complicated but basically just generates the config item name on the fly. No custom function required.

Let me know what you think.

pkg/aws/instance_info_test.go
linki: So it works fine for on-demand pools. The pool will have the first instance type with the correct AMI.

For spot pools it's not perfect yet. It takes the first types AMI but still leaves the all the instance types in the configuration. I guess when the type comes up it won't work. However, this would also happen now if you specify a type that's not compatible.

I would go ahead with this change keeping this in mind for spot pools and improve the UX (e.g. warning or error, docs) in a follow up.

linki: I also added a test for that below.

mikkeloscar:
:+1:

linki:
👍

PR #591, opened 2022-06-27
Introduce sumQuantities template function

mikkeloscar:
This introduces a ~`subQuantity`~ `sumQuantities` template function which can take any number of Quantity type arguments and ~subtract~ sum them ~from each other~.

Example: You have 1 CPU and want to subtract some milicores: `sumQuantities "1" "- 518m" "-10m" = "472m"`

It is needed for this: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5212

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -127,6 +127,7 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"indexedList":                   indexedList,
 		"zoneDistributedNodePoolGroups": zoneDistributedNodePoolGroups,
 		"certificateExpiry":             certificateExpiry,
+		"sumQuantities":                 sumQuantities,
 	}
 
 	content, ok := context.fileData[file]
@@ -674,4 +675,17 @@ func certificateExpiry(certificate string) (string, error) {
 		return "", err
 	}
 	return expiry.UTC().Format(time.RFC3339), nil
+}
+
+func sumQuantities(quantities ...string) (string, error) {
+	var result k8sresource.Quantity
+	for _, quantity := range quantities {
+		q, err := k8sresource.ParseQuantity(quantity)
+		if err != nil {
+			return "", err
+		}
+		result.Add(q)
+	}
+
+	return result.String(), nil
 }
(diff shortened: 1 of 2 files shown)


provisioner/template.go
```diff
@@ -675,3 +676,20 @@ func certificateExpiry(certificate string) (string, error) {
 	}
 	return expiry.UTC().Format(time.RFC3339), nil
 }
+
+func subQuantity(quantities ...string) (string, error) {
+	var result k8sresource.Quantity
```
AlexanderYastrebov, line 681: It looks like (did not check) that Quantity supports sign https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity

If so it might be better to introduce `sumQuantities` to support both addition and subtraction.
`subQuantity("1", "100m", "1m")` would be then `sumQuantities("1", "-100m", "-1m")`

mikkeloscar: I changed it to `subQuantities`. It makes the templating on kubernetes-on-aws side a little more verbose, but seems reasonable when it's more generic.

provisioner/template.go
```diff
@@ -675,3 +676,20 @@ func certificateExpiry(certificate string) (string, error) {
...
 }
+
+func sumQuantities(quantities ...string) (string, error) {
+	var result k8sresource.Quantity
+	for _, quantity := range quantities {
+		q, err := k8sresource.ParseQuantity(quantity)
+		if err != nil {
+			return "", err
+		}
+		if result.IsZero() {
```
AlexanderYastrebov, line 687: minor: `result.IsZero()` seems redundant as result is zero by default and Add should work

provisioner/template.go
AlexanderYastrebov: Changed to `sumQuantities` I guess.

provisioner/template_test.go
```diff
@@ -905,3 +905,38 @@ CWeOoA==
...
+
+func TestSubQuantity(t *testing.T) {
+	for _, tc := range []struct {
+		name     string
+		template string
+		expected string
+	}{
+		{
+			name:     "whole CPU sub fraction",
+			template: `{{ sumQuantities "2" "-256m" }}`,
```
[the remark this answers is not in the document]
mikkeloscar: fixed

provisioner/template.go
mikkeloscar: Fixed

AlexanderYastrebov:
:+1:

mikkeloscar:
:+1:

PR #608, opened 2022-09-01
Extract S3 upload logic and use it also while provisioning etcd.

gargravarr:
This Pull Request (PR) enables uploading the etcd certificates to the specific S3 bucket. I extracted the logic used with node pools to a new common component (`FilesRenderer`) to avoid duplication. I added also tests to verify rendering of the files template, both for etcd and the already existing node pools.

CLM still runs the legacy etcd, so that we can do a smooth transition to the new etcd Amazon Machine Image (AMI). After we verify and replace all etcd clusters with the new AMI, a follow-up PR will disbable and cleanup the legacy code.

In addition, this PR updates the following:
* Updates the Cluster Registry generated client code (in `pkg/cluster_registry`), and
* Updates the linter to `revive` (`golint` was deprecated), and makes the necessary updates to satisfy the new stricter linter rules.


gargravarr:
:+1:

rodrigodelmonte:
:+1:

rodrigodelmonte:
Thanks for also addressing the golint and cluster-registry client.

PR #618, opened 2022-10-07
Support more nodes based on Pod CIDR size

mikkeloscar:
For enabling more than 1019 nodes in our Kubernetes infrastructure we have made it possible to change the default pod cidr from `10.2.0.0/16` to `10.2.0.0/15`: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5405

However we forgot to allow the cluster-autoscaler to scale beyond the 1019 node limit.

This PR fixes this by:
1. Introducing a new function `nodeCIDRMaxNodesPodCIDR` for calculating the number of nodes based on podCIDR mask size, node mask size and potential reserved capacity.
2. "Deprecating" the former version of the function called `nodeCIDRMaxNodes` which had the podCIDR mask size hardcoded.

With this we can change the usage of `nodeCIDRMaxNodes` to `nodeCIDRMaxNodesPodCIDR` here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/87d5861f145b54ed9120f60a299c0d760ac951a4/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml#L54 and eventually remove the function `nodeCIDRMaxNodes` when it's not used anymore.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -118,7 +118,14 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"amiID": func(imageName, imageOwner string) (string, error) {
 			return amiID(context.awsAdapter, imageName, imageOwner)
 		},
-		"nodeCIDRMaxNodes":              nodeCIDRMaxNodes,
+		// TODO: this function is kept for backward compatibility while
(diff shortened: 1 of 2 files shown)


demonCoder95:
👍

PR #620, opened 2022-10-17
Add note about using multiple config sources

demonCoder95:
Add clarification about the way multiple config sources are merged together.

The internal discussion on this can be found in [this](https://chat.google.com/room/AAAAwnCYKwk/pCqAwnfLbpA) chat thread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ production and test clusters, while keeping the manifests readable:
     …
     ```
 
-### A Note on using Multiple Config Sources
+### A Note on Using Multiple Config Sources
 The CLM supports specifying multiple `config-sources` on the command line, see [here](https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/2bb1b5ffd184dbdffefc4c7529671025edf460c4/config/config.go#L92).
 It's important to understand that the order in which these are provided is important. If the same `config-item`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -196,6 +196,22 @@ production and test clusters, while keeping the manifests readable:
...
+earlier source. For instance, consider a CLM deployment that contains the following arguments:
+```yaml
+...
+          - --config-source=source1:git:example.domain-1
+          - --config-source=source2:example.domain-2
+          - --config-source=source3:git:example.domain-3
+...
+```
+Then for any `config-item`, the value from `source3` will be the final value and in case of it existing in the sources specified earlier, the value will be overridden by the value in `source3`.
+This is because the way in which the CLI is implemented, multiple values specified for a CLI flag are appended, in order, to a slice. Later, in the CLM code, when these are [merged](https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/2bb1b5ffd184dbdffefc4c7529671025edf460c4/cmd/clm/main.go#L44), the order of the
```
mikkeloscar, line 212: > This is because the way in which the CLI is implemented

I would either not say this or change it to `this is intended behavior`. It's confusingly detailed if you don't have the context of our internal discussion I think.

demonCoder95: Hmm I agree. I'll change this.

szuecs:
:+1:

PR #621, opened 2022-10-18
template: add filter, ignore and join string helpers

AlexanderYastrebov:


AlexanderYastrebov:
Lets introduce special helpers when needed.

PR #622, opened 2022-10-19
Detect karpenter support and use new manifest apply order

linki:
Changes the apply order based on whether Karpenter is used or not.


linki:
👍

PR #633, opened 2023-01-13
annotate resources for deletion

myaser:
teapot/issues#1950

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -1151,8 +1151,6 @@ func overrideDeletionProtection(ctx context.Context, kind, name string, iface dy
 	switch kind {
 	case "Namespace":
 		annotation = "zalando.org/delete-namespace"
-	case "Postgresql":
-		annotation = "zalando.org/delete-clustername"
 	default:
 		// no annotation needed
 		return nil
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -3,6 +3,7 @@ package provisioner
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -36,6 +37,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1unstructured "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/restmapper"
@@ -1113,8 +1115,11 @@ func performDeletion(ctx context.Context, logger *log.Entry, client dynamic.Inte
 	} else {
 		iface = client.Resource(gvr)
 	}
-
 	if deletion.Name != "" {
+		err := overrideDeletionProtection(ctx, deletion.Kind, deletion.Name, iface)
+		if err != nil {
+			return err
+		}
 		return deleteResource(ctx, iface, logger, deletion.Kind, deletion.Name, deletion.options())
 	} else if len(deletion.Labels) > 0 {
 		items, err := listResources(ctx, iface, deletion)
@@ -1127,6 +1132,10 @@ func performDeletion(ctx context.Context, logger *log.Entry, client dynamic.Inte
 		}
 
 		for _, item := range items {
+			err := overrideDeletionProtection(ctx, deletion.Kind, item.GetName(), iface)
+			if err != nil {
+				return err
+			}
 			err = deleteResource(ctx, iface, logger, deletion.Kind, item.GetName(), deletion.options())
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


provisioner/clusterpy.go
```diff
@@ -1137,6 +1146,38 @@ func performDeletion(ctx context.Context, logger *log.Entry, client dynamic.Inte
 	return nil
 }
 
+func annotateDeletion(ctx context.Context, kind, name string, iface dynamic.ResourceInterface) error {
```
linki, line 1149: Let's rename this so it's easier to understand what this is about. It's some sort of deletion protection, right? So maybe `disableDeletionProtection` or something like that? (similar to CloudFormation's Stack deletion protection)

linki: @myaser Thanks!

myaser:
👍🏻

provisioner/clusterpy.go
```diff
@@ -1137,6 +1146,38 @@ func performDeletion(ctx context.Context, logger *log.Entry, client dynamic.Inte
...
+	default:
+		// no annotation needed
+		return nil
+	}
+
+	patch := map[string]interface{}{
+		"metadata": map[string]interface{}{
+			"annotations": map[string]interface{}{
+				"zalando.org/delete-date": time.Now().Format("2006-01-02"),
+				annotation:                name,
```
linki, line 1165: `name` isn't correct for Postresql, isn't it? Because the annotation name is `"zalando.org/delete-clustername"` the value seems to be the cluster's name and not the database itself.

I also wouldn't mind leaving out Postgresql entirely here. We don't need it right now and this way we cannot accidentally delete it.

provisioner/clusterpy_test.go
```diff
@@ -514,6 +516,40 @@ func TestPerformDeletion(t *testing.T) {
...
+						}
+					}
+				}
+				return []clientError{
+					{
+						matches: deleteAction("namespaces", "ns-foo"),
+						err:     fmt.Errorf("admission webhook \"namespace-admitter.teapot.zalan.do\" denied the request: annotation zalando.org/delete-date not set in manifest to allow resource deletion. See https://cloud.docs.zalando.net/howtos/delete-protection/"),
+					},
+				}
+			},
```
linki, line 552: How come these errors don't change the `expectDeleted` outcome of this test case?

provisioner/clusterpy_test.go
```diff
@@ -697,14 +748,16 @@ func TestPerformDeletion(t *testing.T) {
...
+					for _, ce := range tc.clientErrors(state) {
+						if ce.matches(action) {
+							ignore = true
+							break
+						}
 					}
 				}
 
-				if action, ok := action.(k8stesting.DeleteAction); ok && !ignore {
+				if action, ok := action.(k8stesting.DeleteAction); ok && !ignore && action.GetVerb() == "delete" {
```
linki, line 760: Why is `&& action.GetVerb() == "delete"` needed now? Is it now implied by `action` being a `k8stesting.DeleteAction` already or did something change?

provisioner/clusterpy_test.go
myaser: Since I converted it to a function, I now check if the status map has the delete annotations, then it returns a nil

so, the error isn't actually raised and thus the output does not change

provisioner/clusterpy_test.go
myaser: because the `clientReactors` is adding one more action (`k8stesting.PatchAction`) that is handling the annotation

so the if statement would have caught 2 actions and duplicated the output

> is it now implied by action being a k8stesting.DeleteAction already
it was already the case no change here isn't it?

provisioner/clusterpy.go
myaser: > We don't need it right now and this way we cannot accidentally delete it.

I agree, I will remove it then

provisioner/clusterpy_test.go
linki: I mean, we already check that `action` is a `k8stesting.DeleteAction` (via `ok`). A `k8stesting.PatchAction` should already be filtered out there, shouldn't it?

myaser: I thought so, but no, it is not filtered out.
because of the hierarchy of the Action interfaces and their implementations, the PatchActionImpl is also implementing DeleteAction. 
```
replace (
	k8s.io/apimachinery => k8s.io/apimachinery v0.21.9
	k8s.io/client-go => k8s.io/client-go v0.21.9
)
```
```go
// You can edit this code!
// Click here and start typing.
package main

import (
	"fmt"

	"k8s.io/apimachinery/pkg/runtime/schema"
	k8stypes "k8s.io/apimachinery/pkg/types"
	k8stesting "k8s.io/client-go/testing"
)

func test(a k8stesting.Action) {
	if _, ok := a.(k8stesting.DeleteAction); ok {
		fmt.Println("Action is a DeleteAction")
	} else {
		fmt.Println("Action is not DeleteAction")
	}
}

func main() {
	gvr := schema.GroupVersionResource{Group: "group", Version: "version", Resource: "namespaces"}
	testAction := k8stesting.NewRootPatchAction(gvr, "fake", k8stypes.JSONPatchType, nil)
	test(testAction)
}
```
Screenshot 2023-01-17 at 11 05 34

it seems a bug that exists in this version of k8s, I checked the latest version and it have been fixed

myaser:
👍🏻

provisioner/clusterpy_test.go
linki: 👍

linki: image

linki:
👍

linki:
👍

PR #635, opened 2023-01-23
Fix namespace delete annotation

mikkeloscar:
Follow up to #633 

This changes the patching of Namespace resources to first get the resource, annotate it and update it. The motivation for this is to fix the patching which was not a valid patch resulting in the error:

```
Failed to process cluster: json: cannot unmarshal object into Go value of type jsonpatch.Patch
```

Additionally it _removes_ the existing unit test which attempted to mock the deletion check, but did it in a way, where it didn't correctly test the patch and therefore didn't detect the invalid patch format.

The ideal way to test this would be to mock the delete action to check for the annotation as is done in the admission-controller. However, because delete actions don't include the object content it's very complicated to mock it in a meaningful way.
Therefore leave out the unit test for now.

This is tested manually in a pet cluster.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -3,7 +3,6 @@ package provisioner
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


myaser:
👍🏻

mikkeloscar:
:+1:

PR #636, opened 2023-01-31
Add MemoryFraction function to InstanceInfo

mikkeloscar:
This adds a `MemoryFraction` method which allows parsing a fraction (whole number between `0-100`) and get the percentage of instance memory in bytes returned.

This is useful for calculation e.g. 50% of memory of a master node and set this as the limit for the kube-apiserver pod: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5722

### DIFF
--- a/pkg/aws/instance_info.go
+++ b/pkg/aws/instance_info.go
@@ -35,6 +35,10 @@ func (i Instance) AvailableStorage(instanceStorageScaleFactor float64, rootVolum
 	return int64(instanceStorageScaleFactor * float64(i.InstanceStorageDevices*i.InstanceStorageDeviceSize))
 }
 
+func (i Instance) MemoryFraction(percent int64) int64 {
+	return percent * i.Memory / 100
+}
+
 type InstanceTypes struct {
 	instances map[string]Instance
 }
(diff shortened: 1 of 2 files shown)


pkg/aws/instance_info.go
```diff
@@ -35,6 +35,10 @@ func (i Instance) AvailableStorage(instanceStorageScaleFactor float64, rootVolum
 	return int64(instanceStorageScaleFactor * float64(i.InstanceStorageDevices*i.InstanceStorageDeviceSize))
 }
 
+func (i Instance) MemoryFraction(fraction int64) int64 {
+	return int64(float64(fraction) / 100.0 * float64(i.Memory))
```
linki, line 39: Since input and output are both `int`, there's a nice way to avoid the floats:
```go
func (i Instance) MemoryFraction(percent int64) int64 {
	return percent * i.Memory / 100
}
```

mikkeloscar: Thanks! Changed!

mikkeloscar:
:+1:

mikkeloscar:
:+1:

linki:
👍

PR #643, opened 2023-02-06
Update generated code after go-swagger update

mikkeloscar:
Follow up to #641 where go-swagger was updated which changed some of the generated code for the cluster-registry client.
Without this the files changes during the build and the version gets the `-dirty` suffix.


mikkeloscar:
:+1:

katyanna:
👍

PR #645, opened 2023-02-23
Build multi-arch image for CLM

mikkeloscar:
This adds multi-arch build for CLM to avoid depending on the legacy open source registry.

Further we can use the new dependency support for Zalando ECR based images to trigger the internal deployment.

* It also updates `kubectl` in the image to v1.22.17 as we run that in all clusters by now.


mikkeloscar:
:+1:

katyanna:
👍

PR #653, opened 2023-03-10
refactoring

myaser:
teapot/issues#3254
tested here https://github.com/zalando-incubator/kubernetes-on-aws/pull/5811/files

intuition:
* remove `updatestrategy.NodePoolManager` dependency as it is not used by the provisions
* introduce `NodePoolTemplateRenderer` to simplify the `AWSNodePoolProvisioner` type
* use a map for grouping nodePools instead of slice - refactor `groupNodePools`


provisioner/node_pools.go
```diff
@@ -53,46 +53,111 @@ type remoteData struct {
...
+		return "", err
+	}
+	rendered, err := renderSingleTemplate(cloudInitContents, r.cluster, nodePool, values, r.awsAdapter)
+	if err != nil {
+		return "", err
+	}
 
-	stackManifest, err := p.config.NodePoolManifest(nodePool.Profile, stackFileName)
+	if !r.encodeUserData {
+		return rendered, nil
```
linki, line 111: What's the use case of this? For testing?

I'm asking because according to the [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html) userdata always needs to be base64 encoded, so what's the point of not encoding it?

myaser: I tried to make karpenter work with this way of encoded userdata and I failed 😅 
So as a workaround, I bypassed the encoding, and that worked fine

we will need to revisit this in the future and enforce the encoding (after learning the right way to do it)
but for now, I did not want to be blocked by it

api/node_pool.go
```diff
@@ -25,3 +25,35 @@ func (np NodePool) IsSpot() bool {
...
+				Value:  taintData[1],
+				Effect: taintData[2],
+			})
+		} else if len(taintData) == 2 {
+			taints = append(taints, Taint{
+				Key:    taintData[0],
+				Effect: taintData[1],
+			})
+		}
+	}
```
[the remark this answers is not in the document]
myaser: good catch 😄 

will fix it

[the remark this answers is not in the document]
myaser: in fact the implementation mentioned here https://github.com/kubernetes/kops/blob/7b4430ff61ee497aee0ec28d625f69125e8373ee/pkg/model/awsmodel/spotinst.go#L1070-L1092

suffers the same problem (dedicated:foo=NoSchedule)

I will improve on that

provisioner/node_pools.go
linki: At some point we added the `gzip` part because either the Userdata or the CloudFormation Template size exceeded some limit. I don't know if that's still an issue but just a heads up in case you stumble upon it.

api/node_pool.go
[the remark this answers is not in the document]
myaser: I prevented this case  of `dedicated=test` & `dedicated:foo=NoSchedule`
see pkg/kubernetes/kubernetes_test.go

provisioner/node_pools.go
myaser: Yea, I did stumble upon it, but I could guess it is related to the payload size

In karpenter, userdata is not part of the cloudformation template; rather, it is used by the karpenter controller to create the launch template and then the ec2 node dynamically. Somewhere in this pipeline, it was not able to parse the gzipped-bas64-encoded userdata and was creating corrupted launch templates, so the ec2 node never run

linki: Thanks for the explanation.

provisioner/clusterpy.go
```diff
@@ -321,56 +321,37 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
...
-			default:
-				// update nodes
-				for _, nodePool := range g.NodePools {
-					err := updater.Update(ctx, nodePool)
-					if err != nil {
-						return err
-					}
+	if err = ctx.Err(); err != nil {
+		return err
+	}
```
myaser, line 350: maybe this is redundent

linki:
👍

myaser:
👍🏻

PR #654, opened 2023-03-14
Allow to set apply-only per cluster via config item

linki:
Sometimes you want to have CLM update manifests as fast as possible and skip and defer any node rolls. CLM has a global falg for that `--apply-only` but it's rarely used since it requires to redeploy CLM and it's global.

This adds a config item, similar to `update_strategy`, that allows to config the apply-only behaviour at runtime on a per-cluster basis.

Note: don't merge, untested

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -55,6 +55,7 @@ const (
 	configKeyUpdateStrategy            = "update_strategy"
 	updateStrategyRolling              = "rolling"
 	updateStrategyCLC                  = "clc"
+	configKeyUpdateApplyOnly           = "update_apply_only"
 	defaultMaxRetryTime                = 5 * time.Minute
 	clcPollingInterval                 = 10 * time.Second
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
> // briefly block cluster update to retrigger CLM

Why is this needed? I think CLM will be triggered by the `update_apply_only` config item change, won't it?

linki:
@AlexanderYastrebov Yes, if there's no update in progress. The example specifically handles the case where a long-running cluster update with node rotations is already underway. In this case changing the config item will not retrigger CLM until the node rotations are done which can take a long time. Restarting CLM or blocking/unblocking does the trick then.

linki:
I'm closing this: Since karpenter, worker nodes are rolled _after_ manifests are applied. This significantly reduces the time until manifests are applied and makes this feature mostly unnecessary. We still have to wait for masters to be rolled but that doesn't take long.

PR #655, opened 2023-03-14
deletions: add selector

AlexanderYastrebov:
Adds deletion `selector` to enable set and equality-based label requirements, see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors

It could be used to create resources with dynamic name:
```yaml
# pcs.yaml
# skipper_oauth2_ui_login_hostnames: "foo.example.org=foo,foo.example.com=foo,bar.example.net=bar"

{{ $config_hash := printf "%.32s" (.ConfigItems.skipper_oauth2_ui_login_hostnames | sha256) }}

{{ range split .ConfigItems.skipper_oauth2_ui_login_hostnames "," }}
{{ $hostname := index (split . "=") 0 }}
---
apiVersion: "zalando.org/v1"
kind: PlatformCredentialsSet
metadata:
  name: "{{ $hostname }}-credentials"
  namespace: kube-system
  labels:
    type: hostname-credentials
    config-hash: {{ $config_hash }}
  ...
spec:
...
{{ end }}
```
and delete resources that do not match actual config value hash (note `!=` operator):
```yaml
# deletions.yaml

{{ $config_hash := printf "%.32s" (.ConfigItems.skipper_oauth2_ui_login_hostnames | sha256) }}

post_apply:
- kind: PlatformCredentialsSet
  namespace: kube-system
  selector: type=hostname-credentials,config-hash!={{ $config_hash }}
```

https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/656 proposes to add `sha256` function.

PR #656, opened 2023-03-14
template: add sha256 function

AlexanderYastrebov:
For https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/655

### DIFF
--- a/provisioner/template_test.go
+++ b/provisioner/template_test.go
@@ -104,6 +104,16 @@ func TestManifestHashRecursiveInclude(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestSha256(t *testing.T) {
+	result, err := renderSingle(
+		t,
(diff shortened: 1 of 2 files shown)


provisioner/template_test.go
```diff
@@ -104,6 +104,16 @@ func TestManifestHashRecursiveInclude(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestSha256(t *testing.T) {
+	result, err := renderSingle(
+		t,
+		`{{ printf "%.32s" (.Values.data | sha256) }}`,
```
AlexanderYastrebov, line 110: sha256 is a well-defined function so I propose to implement it as is and derive shorter values if needed.

mikkeloscar:
:+1:

linki:
👍

PR #657, opened 2023-03-27
provisioner: add replaceAll template function

AlexanderYastrebov:
This could be used to define a valid yaml that requires an integer value:
```yaml
spec:
  # {{ replaceAll `
  terminationGracePeriodSeconds: 30
  # ` "30" .Cluster.ConfigItems.skipper_termination_grace_period }}
```
becomes
```yaml
spec:
  # 
  terminationGracePeriodSeconds: 10
  #
```


mikkeloscar:
Closing since it was not used so far

PR #658, opened 2023-03-29
provisioner: update split template function logic for empty input

AlexanderYastrebov:
Splitting empty string intuitively should produce zero parts.

Stdlib strings.Split returns s if s does not contain d therefore strings.Split("", ",") returns one part [""].

This should simplify a common pattern:
```
{{ if ne .ConfigItems.foo "" }}
{{ range $bar := split .ConfigItems.foo "," }}
  - "{{ $bar }}"
{{ end }}
{{ end }}
```
to
```
{{ range $bar := split .ConfigItems.foo "," }}
  - "{{ $bar }}"
{{ end }}
```

### DIFF
--- a/provisioner/template_test.go
+++ b/provisioner/template_test.go
@@ -180,6 +180,38 @@ func TestAZCountNoSubnets(t *testing.T) {
 	require.EqualValues(t, "0", result)
 }
 
+func TestSplit(t *testing.T) {
+	t.Run("empty", func(t *testing.T) {
+		result, err := renderSingle(
+			t,
+			`{{ range $index, $element := split .Values.data "," }}{{ $index }}={{ $element}}{{end}}`,
+			"")
+
+		require.NoError(t, err)
+		require.Equal(t, "", result)
+	})
+
+	t.Run("single", func(t *testing.T) {
+		result, err := renderSingle(
+			t,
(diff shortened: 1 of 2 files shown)


provisioner/template_test.go
```diff
@@ -180,6 +180,38 @@ func TestAZCountNoSubnets(t *testing.T) {
...
+	})
+
+	t.Run("multiple", func(t *testing.T) {
+		result, err := renderSingle(
+			t,
+			`{{ range $index, $element := split .Values.data "," }}{{ $index }}={{ $element}}{{end}}`,
+			"foo,bar")
+
+		require.NoError(t, err)
+		require.Equal(t, "0=foo1=bar", result)
```
RomanZavodskikh, line 211: minor: "0=foo1=bar" this format seems very strange to me. However, I think it already works here, so see no problem merging it.

RomanZavodskikh (approved):

RomanZavodskikh:
:+1:

provisioner/template_test.go
AlexanderYastrebov: This is not a format really but a test template to show what are the values it ranges over

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #692, opened 2023-08-17
add template function

myaser:
Add template function to be used in k8s-on-aws; it is grouping node-pools by the dedicated label, then finding if there is a common profile, it is then return a string value "karpenter", "zalando", or "" based on the profile

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -131,18 +131,19 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"nodeCIDRMaxNodes": func(maskSize int64, reserved int64) (int64, error) {
 			return nodeCIDRMaxNodes(16, maskSize, reserved)
 		},
-		"nodeCIDRMaxNodesPodCIDR":       nodeCIDRMaxNodes,
-		"nodeCIDRMaxPods":               nodeCIDRMaxPods,
-		"parseInt64":                    parseInt64,
-		"generateJWKSDocument":          generateJWKSDocument,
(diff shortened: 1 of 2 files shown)


provisioner/template_test.go
```diff
@@ -1052,3 +1052,115 @@ func TestAWSValidID(t *testing.T) {
...
+						"labels": "dedicated=application-1",
+					},
+				},
+			},
+			expected: map[string]string{
+				"application-1": "zalando",
+			},
+		},
+		{
+			name: "application-1 dedicated pools share the same profile, which is un-known",
```
gargravarr, line 1085: un-known -> unknown

gargravarr:
:+1:

myaser:
👍🏻

provisioner/template.go
```diff
@@ -686,13 +698,34 @@ func zoneDistributedNodePoolGroups(nodePools []*api.NodePool) map[string]bool {
 		}
 	}
 
-	result := make(map[string]bool)
+	return poolGroups
+}
+
+// nodeLifeCycleProviderPerNodePoolGroup groups node-pools by the dedicated label, each node-pool-group is mapped to
+// the provider suitable for its node-pool profile
+// in case all pools of a group do not share the same profile, the group is mapped to empty string
```
mikkeloscar, line 706: What does this mean in practice? Which provider would be used here by admission-controller? :thinking:

myaser: an empty string "" provider will be used, which is practically turning this feature off

PR #693, opened 2023-08-23
Make etcd stack name configurable via config item

mikkeloscar:
This makes the etcd stack name configurable via the `etcd_stack_name` configItem. The purpose of this is to allow setting up a custom etcd cluster for testing.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #551, opened 2022-01-05
Drop OIDC Provider provision code

mikkeloscar:
OIDCProviders can now be provisioned via CloudFormation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-oidcprovider.html

Therefore we can drop the custom provision code from CLM.

Depends on https://github.com/zalando-incubator/kubernetes-on-aws/pull/4861

PR #706, opened 2023-10-11
all: remove deadcode

AlexanderYastrebov:
* getEnvVars usage removed in #573
* getHostname usage removed in #551

Found via `deadcode -test ./...`

See https://pkg.go.dev/golang.org/x/tools/internal/cmd/deadcode


katyanna:
👍

mikkeloscar:
:+1:

PR #708, opened 2023-10-20
provisioner: add json template function

AlexanderYastrebov:
This change adds `json` function that enables use of json-formatted strings in templates.

This could be used to define structured configuration values or repeated chunks in templates that require modification of several values, see provisiner/testdata/json.template.yaml for an example use.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/binary"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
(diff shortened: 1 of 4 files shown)


AlexanderYastrebov:
JSON allows arbitrary tree-like data and requires user to be careful with quoting and commas. For templating the simple and better approach seems to be a `dict` constructor function https://stackoverflow.com/questions/18276173/calling-a-template-with-several-pipeline-parameters/18276968#18276968

AlexanderYastrebov:
See https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/711

PR #711, opened 2023-11-29
provisioner: add dict template function

AlexanderYastrebov:
Add dict template function inspired by
https://stackoverflow.com/questions/18276173/calling-a-template-with-several-pipeline-parameters/18276968#18276968

Compared to stackoverflow version it additionally checks for duplicate keys.

Also decided not to use `dict` from github.com/Masterminds/sprig/v3 as it allows non-string keys, uneven number of arguments and does not check for duplicate keys.


AlexanderYastrebov:
Here is the prospective usecase https://github.com/zalando-incubator/kubernetes-on-aws/pull/6592

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #712, opened 2023-11-29
all: fix golangci-lint errors

AlexanderYastrebov:
Fixes "indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (revive)"

### DIFF
--- a/pkg/aws/instance_info.go
+++ b/pkg/aws/instance_info.go
@@ -124,34 +124,33 @@ func (types *InstanceTypes) SyntheticInstanceInfo(instanceTypes []string) (Insta
 		return Instance{}, fmt.Errorf("no instance types provided")
 	} else if len(instanceTypes) == 1 {
 		return types.InstanceInfo(instanceTypes[0])
-	} else {
-		first, err := types.InstanceInfo(instanceTypes[0])
+	}
+
+	first, err := types.InstanceInfo(instanceTypes[0])
+	if err != nil {
+		return Instance{}, err
+	}
+
+	result := Instance{
+		InstanceType:              "<multiple>",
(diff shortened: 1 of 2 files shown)


pkg/aws/instance_info.go
```diff
@@ -124,34 +124,34 @@ func (types *InstanceTypes) SyntheticInstanceInfo(instanceTypes []string) (Insta
...
-			result.InstanceStorageDevices = min(result.InstanceStorageDevices, info.InstanceStorageDevices)
-		}
-		result.InstanceType = "<multiple>"
-		return result, nil
+		result.VCPU = min(result.VCPU, info.VCPU)
+		result.Memory = min(result.Memory, info.Memory)
+		result.InstanceStorageDeviceSize = min(result.InstanceStorageDeviceSize, info.InstanceStorageDeviceSize)
+		result.InstanceStorageDevices = min(result.InstanceStorageDevices, info.InstanceStorageDevices)
 	}
+	result.InstanceType = "<multiple>"
```
szuecs, line 153: So if there is only one instance type, we have now `<multiple>`

szuecs: ah no in line 125 we check that we have one and return early. This means you can move result.InstanceType to the top.

AlexanderYastrebov: There is a check `if len(instanceTypes) == 1` above. The change is equivalent as it just removes else branch where other if branches have return

AlexanderYastrebov: Fixed

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #722, opened 2023-12-07
controller: compact repeating problems

AlexanderYastrebov:
While problem list is limited to 25 items the error itself is often re-occuring so it makes sense to compact same subsequent errors into one.

- [ ] requites https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/721 due to use of new `slices` package


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #732, opened 2024-01-18
Drop legacy etcd support

mikkeloscar:
This drops the legacy etcd hacks which are no longer needed.

Ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/6756


gargravarr:
:+1:

PR #743, opened 2024-02-01
add template functions for dynamic resource scaling

demonCoder95:
Adds the following template functions:

1. `scaleQuantity` - which scales a Kubernetes [resource.Quantity](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity) type represented as a `string` by a `float32` factor.
2. `instanceTypeCPUQuantity` - which returns vCPU count of an EC2 instance as Kubernetes resource.Quantity type represented as a `string`.
3. `instanceTypeMemoryQuantity` - which returns memory (in MiB) of an EC2 instance as Kubernetes resource.Quantity type represented as a `string`.

Also does the following refactoring:

1. Calls `NewInstanceTypesFromAWS` from `prepareProvision` instead of `Provision` function and it also returns the same to be used in `Provision` as well. This is a single in-memory representation of `instanceTypes` data that will be shared everywhere.
2. Extends `renderSingleTemplate` to receive the in-memory data for `instanceTypes` used to create the `TemplateContext`.
3. Extends `renderTemplate` to receive the same `instanceTypes` data.
4. Updates all instances in the code that use `renderTemplate` and `renderSingleTemplate` functions to ensure they receive the right data.
5. Moves `instanceTypes` field from `AWSNodePoolProvisioner` to `NodePoolTemplateRenderer` so both provisioners can use the same data in their rendering functions.
6. Extends `FilesRenderer` to include `instanceTypes` field that is passed onto rendering functions. This data is received from the `Provision` function.

The motivation is to be able to slice instance resources of an instance type given. This will then be used to size VPA limits for daemonsets. This is especially useful for master node daemonset VPA sizing in test clusters where we have very small instances where some pods cannot run sometimes because of overcommitted CPU.

Addresses Issue: https://github.bus.zalan.do/teapot/issues/issues/3469

The supporting PR in `kubernetes-on-aws` which demonstrates the usage of these new template functions: https://github.com/zalando-incubator/kubernetes-on-aws/pull/6931

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -106,7 +106,7 @@ func (p *clusterpyProvisioner) Supports(cluster *api.Cluster) bool {
 	return cluster.Provider == providerID
 }
 
-func (p *clusterpyProvisioner) updateDefaults(cluster *api.Cluster, channelConfig channel.Config, adapter *awsAdapter) error {
+func (p *clusterpyProvisioner) updateDefaults(cluster *api.Cluster, channelConfig channel.Config, adapter *awsAdapter, instanceTypes *awsUtils.InstanceTypes) error {
 	defaultsFiles, err := channelConfig.DefaultsManifests()
 	if err != nil {
 		return err
@@ -118,7 +118,7 @@ func (p *clusterpyProvisioner) updateDefaults(cluster *api.Cluster, channelConfi
 	allDefaults := make(map[string]string)
 
 	for _, file := range defaultsFiles {
-		result, err := renderSingleTemplate(file, &withoutConfigItems, nil, nil, adapter)
+		result, err := renderSingleTemplate(file, &withoutConfigItems, nil, nil, adapter, instanceTypes)
 		if err != nil {
 			return err
 		}
@@ -176,7 +176,7 @@ func (p *clusterpyProvisioner) propagateConfigItemsToNodePools(cluster *api.Clus
 // Provision provisions/updates a cluster on AWS. Provision is an idempotent
 // operation for the same input.
 func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry, cluster *api.Cluster, channelConfig channel.Config) error {
-	awsAdapter, updater, err := p.prepareProvision(logger, cluster, channelConfig)
+	instanceTypes, awsAdapter, updater, err := p.prepareProvision(logger, cluster, channelConfig)
 	if err != nil {
 		return err
 	}
@@ -271,11 +271,12 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 	}
 
 	// render the manifests to find out if they're valid
-	deletions, err := parseDeletions(channelConfig, cluster, values, awsAdapter)
+	deletions, err := parseDeletions(channelConfig, cluster, values, awsAdapter, instanceTypes)
 	if err != nil {
 		return err
 	}
-	manifests, err := renderManifests(channelConfig, cluster, values, awsAdapter)
+
+	manifests, err := renderManifests(channelConfig, cluster, values, awsAdapter, instanceTypes)
 	if err != nil {
 		return err
 	}
@@ -290,7 +291,7 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 
 	// create or update the etcd stack
 	if p.manageEtcdStack {
-		err = createOrUpdateEtcdStack(ctx, logger, channelConfig, cluster, values, etcdKMSKeyARN, awsAdapter, bucketName)
+		err = createOrUpdateEtcdStack(ctx, logger, channelConfig, cluster, values, etcdKMSKeyARN, awsAdapter, bucketName, instanceTypes)
 		if err != nil {
 			return err
 		}
@@ -300,7 +301,7 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 		return err
 	}
 
-	outputs, err := createOrUpdateClusterStack(ctx, channelConfig, cluster, values, awsAdapter, bucketName)
+	outputs, err := createOrUpdateClusterStack(ctx, channelConfig, cluster, values, awsAdapter, bucketName, instanceTypes)
 	if err != nil {
 		return err
 	}
@@ -308,11 +309,6 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 
 	if err = ctx.Err(); err != nil {
 		return err
-	}
-
-	instanceTypes, err := awsUtils.NewInstanceTypesFromAWS(awsAdapter.ec2Client)
-	if err != nil {
-		return fmt.Errorf("failed to fetch instance types from AWS")
 	}
 
 	// provision node pools
@@ -324,9 +320,9 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 			bucketName:     bucketName,
 			logger:         logger,
 			encodeUserData: true,
+			instanceTypes:  instanceTypes,
 		},
-		instanceTypes: instanceTypes,
-		azInfo:        azInfo,
+		azInfo: azInfo,
 	}
 
 	karpenterProvisioner, err := NewKarpenterNodePoolProvisioner(
@@ -337,6 +333,7 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 			bucketName:     bucketName,
 			logger:         logger,
 			encodeUserData: false,
+			instanceTypes:  instanceTypes,
 		}, p.execManager, p.tokenSource,
 	)
 	if err != nil {
@@ -405,6 +402,7 @@ func createOrUpdateEtcdStack(
 	etcdKmsKeyARN string,
 	adapter *awsAdapter,
 	bucketName string,
+	instanceTypes *awsUtils.InstanceTypes,
 ) error {
 
 	template, err := config.EtcdManifest(etcdStackFileName)
@@ -419,11 +417,12 @@ func createOrUpdateEtcdStack(
 	}
 
 	renderer := &FilesRenderer{
-		awsAdapter: adapter,
-		cluster:    cluster,
-		config:     config,
-		directory:  "etcd",
-		nodePool:   nil,
+		awsAdapter:    adapter,
+		cluster:       cluster,
+		config:        config,
+		directory:     "etcd",
+		nodePool:      nil,
+		instanceTypes: instanceTypes,
 	}
 
 	s3Path, err := renderer.RenderAndUploadFiles(values, bucketName, etcdKmsKeyARN)
@@ -434,7 +433,7 @@ func createOrUpdateEtcdStack(
 	logger.Debugf("Uploaded generated files to %s", s3Path)
 	values[s3GeneratedFilesPathValuesKey] = s3Path
 
-	rendered, err := renderSingleTemplate(template, cluster, nil, values, adapter)
+	rendered, err := renderSingleTemplate(template, cluster, nil, values, adapter, instanceTypes)
 	if err != nil {
 		return err
 	}
@@ -478,13 +477,13 @@ func createOrUpdateEtcdStack(
 	return nil
 }
 
-func createOrUpdateClusterStack(ctx context.Context, config channel.Config, cluster *api.Cluster, values map[string]interface{}, adapter *awsAdapter, bucketName string) (map[string]string, error) {
+func createOrUpdateClusterStack(ctx context.Context, config channel.Config, cluster *api.Cluster, values map[string]interface{}, adapter *awsAdapter, bucketName string, instanceTypes *awsUtils.InstanceTypes) (map[string]string, error) {
 	template, err := config.StackManifest(clusterStackFileName)
 	if err != nil {
 		return nil, err
 	}
 
-	rendered, err := renderSingleTemplate(template, cluster, nil, values, adapter)
+	rendered, err := renderSingleTemplate(template, cluster, nil, values, adapter, instanceTypes)
 	if err != nil {
 		return nil, err
 	}
@@ -779,26 +778,32 @@ func (p *clusterpyProvisioner) setupAWSAdapter(logger *log.Entry, cluster *api.C
 // prepareProvision checks that a cluster can be handled by the provisioner and
 // prepares to provision a cluster by initializing the aws adapter.
 // TODO: this is doing a lot of things to glue everything together, this should be refactored.
-func (p *clusterpyProvisioner) prepareProvision(logger *log.Entry, cluster *api.Cluster, channelConfig channel.Config) (*awsAdapter, updatestrategy.UpdateStrategy, error) {
+func (p *clusterpyProvisioner) prepareProvision(logger *log.Entry, cluster *api.Cluster, channelConfig channel.Config) (*awsUtils.InstanceTypes, *awsAdapter, updatestrategy.UpdateStrategy, error) {
 	if cluster.Provider != providerID {
-		return nil, nil, ErrProviderNotSupported
+		return nil, nil, nil, ErrProviderNotSupported
 	}
 
 	logger.Infof("clusterpy: Prepare for provisioning cluster %s (%s)..", cluster.ID, cluster.LifecycleStatus)
 
 	adapter, err := p.setupAWSAdapter(logger, cluster)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to setup AWS Adapter: %v", err)
+		return nil, nil, nil, fmt.Errorf("failed to setup AWS Adapter: %v", err)
 	}
 
-	err = p.updateDefaults(cluster, channelConfig, adapter)
+	// fetch instance data that will be used by all the render functions
+	instanceTypes, err := awsUtils.NewInstanceTypesFromAWS(adapter.ec2Client)
 	if err != nil {
-		return nil, nil, fmt.Errorf("unable to read configuration defaults: %v", err)
+		return nil, nil, nil, fmt.Errorf("failed to fetch instance types from AWS")
+	}
+
+	err = p.updateDefaults(cluster, channelConfig, adapter, instanceTypes)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("unable to read configuration defaults: %v", err)
 	}
 
 	err = p.decryptConfigItems(cluster)
 	if err != nil {
-		return nil, nil, fmt.Errorf("unable to decrypt config items: %v", err)
+		return nil, nil, nil, fmt.Errorf("unable to decrypt config items: %v", err)
 	}
 
 	p.propagateConfigItemsToNodePools(cluster)
@@ -821,7 +826,7 @@ func (p *clusterpyProvisioner) prepareProvision(logger *log.Entry, cluster *api.
 
 	client, err := kubernetes.NewClient(cluster.APIServerURL, p.tokenSource)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	// setup updater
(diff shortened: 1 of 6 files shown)


gargravarr:
Please add some unit tests.

demonCoder95:
Yes, will add them as the next step.

provisioner/template.go
```diff
@@ -733,3 +734,49 @@ func sumQuantities(quantities ...string) (string, error) {
...
+	}
+
+	// validate the resourceType
+	if resourceType != "cpu" && resourceType != "memory" {
+		return 0, fmt.Errorf("resourceType must be either 'cpu' or 'memory'")
+	}
+
+	// get the instance type info
+	input := ec2.DescribeInstanceTypesInput{InstanceTypes: awsUtil.StringSlice([]string{instanceType})}
+	output, err := adapter.ec2Client.DescribeInstanceTypes(&input)
```
AlexanderYastrebov, line 765: How often instance type configurations change? I'd assume its pretty-much constant so maybe, to avoid AWS API call on each template rendering its better to hardcode them or load all types once on CLM startup.

mikkeloscar: I'm sure we have an in-memory representation of instance-types already in CLM, we use that in NodePool configuration iirc.

demonCoder95: You mean something like this here?
https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/master/provisioner/node_pools.go#L308

provisioner/template.go
```diff
@@ -733,3 +734,49 @@ func sumQuantities(quantities ...string) (string, error) {
...
 	return strings.Replace(id, ":", "__", -1)
 }
+
+// divideInstanceResourcesInRatio takes in an instance type and a ratio and returns the result of applying
+// the ratio on the instance CPU and Memory sizes.
+// The ratio must be a positive number and less than 1.
+// The resourceType must be either 'cpu' or 'memory'.
+// The result is rounded down to the nearest integer, to avoid overcommitting resources.
+// The units for CPU and memory are millicores and mebibytes (MiB) respectively.
+func divideInstanceResourcesInRatio(adapter *awsAdapter, instanceType string, resourceType string, ratio float32) (dividedResource int, err error) {
```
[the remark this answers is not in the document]
demonCoder95: I'll check. Thanks!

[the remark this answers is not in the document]
AlexanderYastrebov: Please also consider adding a set of functions:
1. a pair of `instanceCpu` and `instanceMemory` receiving single instance type argument and returns cpu/memory as a k8s quantity
2. `scaleQuantity(value, factor)` returns value*factor as a k8s quantity.

This way we get a set of composable functions that would allow e.g. original scenario via:
```
scaleQuantity (instanceCpu "m4.xlarge") 0.9
```
or fixed-size offset using existing `sumQuantities`:
```
sumQuantities (instanceMemory "m4.xlarge") "-200Mi"
```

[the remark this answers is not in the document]
demonCoder95: I've added the functions in the latest commit, haven't replaced the AWS API calls with the in-memory data yet.

demonCoder95:
Updated PR title and description to better reflect the work.

provisioner/template.go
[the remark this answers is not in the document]
demonCoder95: Also, I think as a template function, `scaleQuantity` should return the `Quantity` as `String` no? So that it can be used directly in the template. I'm basing this on the fact that `sumQuantities` does the same.

provisioner/template.go
demonCoder95: I see we populate the `InstanceTypes` data in `clusterpy.go` [here](https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/master/provisioner/clusterpy.go#L313) and then pass it on to the new `caNodePoolProvisioner` struct [here](https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/master/provisioner/clusterpy.go#L319). I'll figure out a way to use this in-memory representation in `template.go` functions as well. Will have to refactor some parts probably.

demonCoder95:
Added a commit for PoC of refactoring to use the in-memory instance data  in the beginning when CLM starts. This is just to highlight the approach I have in mind for making the data available to the template rendering functions.

provisioner/template.go
demonCoder95: I attempted one way I found of making this happen in the commit I pushed [here](https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/743/commits/ea342af2f9ce76876183c73d5535a1b8ec3d5093).

provisioner/clusterpy.go
```diff
@@ -310,11 +311,6 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 		return err
 	}
 
-	instanceTypes, err := awsUtils.NewInstanceTypesFromAWS(awsAdapter.ec2Client)
```
demonCoder95, line 313: This I moved into the `prepareProvision` function. The data is used in that function and returned as well so it can be used in `Provision` function as well. This way we fetch the data just once on every call to `Provision`.

provisioner/clusterpy.go
```diff
@@ -324,9 +320,9 @@ func (p *clusterpyProvisioner) Provision(ctx context.Context, logger *log.Entry,
 			bucketName:     bucketName,
 			logger:         logger,
 			encodeUserData: true,
+			instanceTypes:  instanceTypes,
```
demonCoder95, line 323: I moved the `instanceTypes` field from `AWSNodePoolProvisioner` to here, where it is more generalized. This way the same field can be used in the render functions for `KarpenterNodePoolProvisioner` as well.

provisioner/template.go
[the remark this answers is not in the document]
AlexanderYastrebov: I think all of them should work with (receive as args and return) string representation of quantity. This would allow passing string const string arguments e.g. `scaleQuantity "100Mi" 0.7`

The cost of parsing it internally should be negligible.

demonCoder95:
Update PR description with details of all the refactoring that has been done.

provisioner/template.go
demonCoder95: Just finished all the refactoring and updated the commit.

provisioner/template.go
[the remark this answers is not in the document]
demonCoder95: I agree. I'll update them to string representations. Then we should be done. Will then add unit tests and that should be it.

provisioner/clusterpy.go
```diff
@@ -106,7 +106,7 @@ func (p *clusterpyProvisioner) Supports(cluster *api.Cluster) bool {
 	return cluster.Provider == providerID
 }
 
-func (p *clusterpyProvisioner) updateDefaults(cluster *api.Cluster, channelConfig channel.Config, adapter *awsAdapter) error {
+func (p *clusterpyProvisioner) updateDefaults(cluster *api.Cluster, channelConfig channel.Config, adapter *awsAdapter, instanceTypes *awsUtils.InstanceTypes) error {
```
AlexanderYastrebov, line 109: I think `InstanceInfo` could be a method of awsAdapter (it can cache result of API call if needed).

Also I now see there is a whole `type Instance` that provides access to instance properties with similar accessors like `MemoryFraction`, see https://github.com/zalando-incubator/kubernetes-on-aws/blob/6f3c355069f1aebda0f77de6d831612f7d33c001/cluster/node-pools/master-default/userdata.yaml#L206

I think we should unify it, maybe by providing all instance types in the values and re-using existing accessors (add another for CPU):
```yaml
memory: {{ .Values.InstanceInfos["c6g-xlarge"].MemoryFraction 70 }}
```
or migrate `.Values.InstanceInfo.MemoryFraction` to proposed new template functions.

provisioner/template.go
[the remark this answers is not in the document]
demonCoder95: Update in the latest commit.

provisioner/clusterpy.go
mikkeloscar: I don't have a strong opinion on this. In the past we just needed to access the current instance in the context of the node pool, so no need to make it more elaborate. Now that we have more use cases I think it's fine to change it. Just keep in mind the migration path, we need to support both until the equivalent kubernetes-on-aws changes would be merged to all channels.

demonCoder95: Yes, I had something similar in mind. We can add the new functions first, roll them out and use them in one VPA to test. Once it runs fine through all channels I can create a follow-up ticket to migrate to the new functions and once the migration finishes, we can remove the old ones.

demonCoder95: I created a new issue for this specific problem. The change in this PR is already quite sizeable and I suggest we move on with this change, propagate it through the channels and then once it's finished, we can move towards the refactoring part.

I'll resolve this conversation now, we can carry on in the new issue, on this topic: https://github.bus.zalan.do/teapot/issues/issues/3566

I'll just proceed to test the feature-complete change in the current state so we can move forward.

provisioner/template.go
```diff
@@ -733,3 +745,50 @@ func sumQuantities(quantities ...string) (string, error) {
...
+// instanceTypeMemory returns the memory of an instance type provided as k8sresource.Quantity represented as string
+func instanceTypeMemory(context *templateContext, instanceType string) (string, error) {
+	// get the instance type info
+	instanceTypeInfo, err := context.instanceTypes.InstanceInfo(instanceType)
+
+	if err != nil {
+		return "", err
+	}
+
+	memory := fmt.Sprintf("%v%s", instanceTypeInfo.Memory/giga, "Gi")
```
AlexanderYastrebov, line 772: I think you can use https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#NewQuantity here
```go
k8sresource.NewQuantity(instanceTypeInfo.Memory, k8sresource.BinarySI)
```

demonCoder95: Ah, that's much better. Thanks 🙏  Added it in the next commit.

demonCoder95: @AlexanderYastrebov do you know of a good way to be able to return the quantity in a "friendly" way?

So, in case of memory, say I have a case of `16Gi` and I do `scaleQuantity "16Gi" 0.1`. I expect the returned quantity to look a more friendly `1.6Gi` instead of `1717986918` bytes. I'm unable to be able to use anything in the `resource` module for this. `quantity.String()` always prints the bytes, which might look ugly in manifests, no?

demonCoder95: I see in our existing method like `MemoryFraction` we return memory in bytes, so maybe bytes is acceptable?
https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/master/pkg/aws/instance_info_test.go#L462

demonCoder95: I also see when manifests are stored in etcd in k8s, the representation of memory is in `bytes`. So, I think we can keep the bytes values for now. I'll mark this as resolved now.

demonCoder95:
I just tested the new functions in a pet cluster and they work great! 🎉 

Used in a VPA resource
```
...
        minAllowed:
          memory: {{ scaleQuantity (instanceTypeMemory "m4.xlarge") 0.1 }}
          cpu: {{ scaleQuantity (instanceTypeCPU "m4.xlarge") 0.1 }}
...
```

And in the cluster the resource created had
```
...
      Min Allowed:
        Cpu:     400m
        Memory:  1717986944
...
```

demonCoder95:
PR has been tested and has all the unit tests also complete. Can be reviewed finally and merged. Also created a proof-of-concept PR in `kubernetes-on-aws` to show how the new template functions will be used.
https://github.com/zalando-incubator/kubernetes-on-aws/pull/6931

provisioner/template.go
```diff
@@ -124,6 +127,13 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"awsValidID":                            awsValidID,
 		"indent":                                sprig.GenericFuncMap()["indent"],
 		"dict":                                  dict,
+		"scaleQuantity":                         scaleQuantity,
+		"instanceTypeCPU": func(instanceType string) (string, error) {
+			return instanceTypeCPU(context, instanceType)
+		},
+		"instanceTypeMemory": func(instanceType string) (string, error) {
+			return instanceTypeMemory(context, instanceType)
+		},
```
AlexanderYastrebov, line 136: Maybe add `Quantity` suffix to highlight the format, i.e. `instanceTypeCPUQuantity` and `instanceTypeMemoryQuantity`

provisioner/template.go
```diff
@@ -733,3 +743,55 @@ func sumQuantities(quantities ...string) (string, error) {
...
+	quantity, err := k8sresource.ParseQuantity(quantityStr)
+	if err != nil {
+		return quantityStr, fmt.Errorf("failed to parse %v as k8sresource.Quantity: %v", quantityStr, err)
+	}
+	// CPU quantities can be represented as milli-cores, so we need to scale them differently
+	if quantity.Value() < 1024 && quantity.Format == k8sresource.DecimalSI {
+		quantity.SetMilli(int64(float32(quantity.MilliValue()) * factor))
+	} else {
+		// memory quantities can be represented as binary SI, so we need to scale them differently
+		scaledValue := int64(float32(quantity.Value()) * factor)
```
AlexanderYastrebov, line 793: Maybe https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity.AsApproximateFloat64 fits better here.

quantity.Value() rounds to integer, e.g. I am not sure if it will do the right thing for `scaleQuantity "2.5Gi", 1`- it should retrurn 2.5Gi (maybe as bytes) but I think it may round 2.5 to 2 first.

provisioner/template.go
```diff
@@ -733,3 +743,55 @@ func sumQuantities(quantities ...string) (string, error) {
...
+// instanceTypeCPU returns the vCPUs of an instance type provided as k8sresource.Quantity represented as string
+func instanceTypeCPU(context *templateContext, instanceType string) (string, error) {
+	// get the instance type info
+	instanceTypeInfo, err := context.instanceTypes.InstanceInfo(instanceType)
+
+	if err != nil {
+		return "", err
+	}
+
+	cpu := fmt.Sprintf("%v", instanceTypeInfo.VCPU)
```
mikkeloscar, line 756: nit: should this just be `%s`?

provisioner/template.go
mikkeloscar: Let's add a test for this @demonCoder95

demonCoder95: That's a good catch. It in fact returns `2560Mi` instead of `2.5Gi`. It didn't at least round it to `2Gi`. I'm thinking when `Gi` units become fractions, it reduces the answer to `Mi`. Is that okay?

provisioner/template.go
demonCoder95: Fixed in the next commit.

demonCoder95: Actually, the linter started complaining that `int64` shouldn't be put where `String` is expected.

provisioner/template.go
mikkeloscar: It makes sense to show `Mi` instead of `Gi`

provisioner/template.go
mikkeloscar: Ah, then it's `%d` :)

provisioner/template.go
demonCoder95: Using `AsApproximateFloat64` didn't change this result. The result is the same as I described above. There is no rounding off in terms of `Gi` units but some rounding errors when going from `Gi` -> `Mi` units.

demonCoder95: Actually, nevermind, 2.5Gi -> 2048Mi + 512Mi -> 2560Mi. So, it's doing correctly 😁  I think it's good! I added the test case in the next commit.

AlexanderYastrebov: BTW, I'd suggest to always prefer float64 unless its really math performance-critical part (i.e. never in our domain).
Float numbers represent integers in certain range without loss of precision, for float32 the limit is quite low, see https://stackoverflow.com/a/3793950/828328

demonCoder95:
There seem to be a lot of linter errors, I'll fix them in the next commit to fix the pipeline.
```
pkg/updatestrategy/aws_asg.go:308:83: unused-parameter: parameter 'lastPage' seems to be unused, consider removing or renaming it as _ (revive)
		err := n.ec2Client.DescribeTagsPages(params, func(resp *ec2.DescribeTagsOutput, lastPage bool) bool {
		                                                                                ^
pkg/updatestrategy/aws_ec2.go:246:2: var-naming: var instanceIds should be instanceIDs (revive)
	instanceIds := make([]*string, 0, len(instances))
	^
pkg/updatestrategy/aws_asg.go:445:116: unused-parameter: parameter 'lastPage' seems to be unused, consider removing or renaming it as _ (revive)
	err := n.asgClient.DescribeAutoScalingGroupsPages(params, func(resp *autoscaling.DescribeAutoScalingGroupsOutput, lastPage bool) bool {
	                                                                                                                  ^
pkg/updatestrategy/aws_asg.go:538:102: unused-parameter: parameter 'lastPage' seems to be unused, consider removing or renaming it as _ (revive)
	err := n.ec2Client.DescribeInstancesPages(describeParams, func(output *ec2.DescribeInstancesOutput, lastPage bool) bool {
	                                                                                                    ^
pkg/updatestrategy/node_pool_manager_test.go:371:18: unused-parameter: parameter 'ctx' seems to be unused, consider removing or renaming it as _ (revive)
	evictPod = func(ctx context.Context, client kubernetes.Interface, logger *log.Entry, pod v1.Pod) error {
	                ^
pkg/updatestrategy/drain_test.go:40:68: unused-parameter: parameter 'logger' seems to be unused, consider removing or renaming it as _ (revive)
	evictPod = func(ctx context.Context, client kubernetes.Interface, logger *log.Entry, pod v1.Pod) error {
	                                                                  ^
pkg/updatestrategy/drain_test.go:202:19: unused-parameter: parameter 'ctx' seems to be unused, consider removing or renaming it as _ (revive)
		evictPod = func(ctx context.Context, client kubernetes.Interface, logger *log.Entry, pod v1.Pod) error {
		                ^
pkg/updatestrategy/drain_test.go:225:69: unused-parameter: parameter 'logger' seems to be unused, consider removing or renaming it as _ (revive)
		evictPod = func(ctx context.Context, client kubernetes.Interface, logger *log.Entry, pod v1.Pod) error {
		                                                                  ^
pkg/updatestrategy/drain_test.go:274:20: unused-parameter: parameter 'ctx' seems to be unused, consider removing or renaming it as _ (revive)
		deletePod = func(ctx context.Context, client kubernetes.Interface, logger *log.Entry, pod v1.Pod) error {
09/02/2024 14:31:07 		                 ^
provisioner/remote_files_test.go:175:26: unused-parameter: parameter 'tt' seems to be unused, consider removing or renaming it as _ (revive)
		t.Run(tc.Message, func(tt *testing.T) {
		                       ^
provisioner/clusterpy_test.go:501:74: unused-parameter: parameter 'r' seems to be unused, consider removing or renaming it as _ (revive)
			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			                                                                      ^
provisioner/clusterpy_test.go:149:4: var-naming: struct field subnetIds should be subnetIDs (revive)
			subnetIds       []string
			^
provisioner/clusterpy.go:217:5: var-naming: var subnetIds should be subnetIDs (revive)
	if subnetIds, ok := cluster.ConfigItems[subnetsConfigItemKey]; ok {
	   ^
controller/cluster_list.go:97:2: var-naming: var availableClusterIds should be availableClusterIDs (revive)
	availableClusterIds := make(map[string]bool)
	^
controller/cluster_list_test.go:116:6: var-naming: func allClusterIds should be allClusterIDs (revive)
func allClusterIds(clusterList *ClusterList) []string {
     ^

```

demonCoder95:
It appears those errors are red herrings, the issue was in the unit test naming after refactoring function names. Fixed in the latest push. Let's see.

demonCoder95:
I just rebased on latest `master`, maybe that will help.

demonCoder95:
Not helping, I'll just fix linter errors in a separate PR then.

demonCoder95:
Fixing linter errors here: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/747

demonCoder95:
Rebasing on latest master after linter fix was merged.

provisioner/template.go
demonCoder95: Added!

demonCoder95:
👍

PR #744, opened 2024-02-01
update karpenter manifest management

myaser:
CLM had the CRDs for karpenter hardcoded.
these strings are used to query k8s api server searching for node pools to be cleaned up
the karpenter API is now upgraded from alpha to beta and thus the CRDs have been renamed and their structure have slightly changed

in this PR is adding support for both versions of karpenter (depends on a config_item)

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -203,12 +203,16 @@ func (c *ClientsCollection) Get(ctx context.Context, kind, namespace, name strin
 	return client.Get(ctx, name, options, subresources...)
 }
 
-func (c *ClientsCollection) Exists(ctx context.Context, kind, namespace, name string, options metav1.GetOptions, subresources ...string) bool {
+func (c *ClientsCollection) Exists(ctx context.Context, kind, namespace, name string, options metav1.GetOptions, subresources ...string) (bool, error) {
 	_, err := c.Get(ctx, kind, namespace, name, options, subresources...)
-	if err != nil { // TODO: check for specific error
-		return false
+
+	if apierrors.IsNotFound(err) {
+		return false, nil
 	}
-	return true
+	if err != nil {
+		return false, err
+	}
+	return true, nil
 }
 
 func (c *ClientsCollection) List(ctx context.Context, kind, namespace string, opts metav1.ListOptions) (*unstructured.UnstructuredList, error) {
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -203,6 +203,14 @@ func (c *ClientsCollection) Get(ctx context.Context, kind, namespace, name strin
 	return client.Get(ctx, name, options, subresources...)
 }
 
+func (c *ClientsCollection) Exists(ctx context.Context, kind, namespace, name string, options metav1.GetOptions, subresources ...string) bool {
+	_, err := c.Get(ctx, kind, namespace, name, options, subresources...)
+	if err != nil { // TODO: check for specific error
+		return false
+	}
+	return true
+}
+
 func (c *ClientsCollection) List(ctx context.Context, kind, namespace string, opts metav1.ListOptions) (*unstructured.UnstructuredList, error) {
 	client, err := c.getResourceClient(kind, namespace)
 	if err != nil {
--- a/pkg/updatestrategy/aws_ec2.go
+++ b/pkg/updatestrategy/aws_ec2.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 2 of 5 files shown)


provisioner/clusterpy.go
```diff
@@ -849,8 +849,9 @@ func (p *clusterpyProvisioner) prepareProvision(logger *log.Entry, cluster *api.
 	if err != nil {
 		return nil, nil, err
 	}
+	crdResolver := NewKarpenterCRDResolver(cluster.ConfigItems["karpenter_main_crd_name"])
```
mikkeloscar, line 852: What if we simply try the newest CRD first and fallback to the other one if it returns not found? Then we don't need to set a config-item and clean that up later and so on?

myaser: after a second thought, I think it is better to inspect the cluster and automatically pick the right CRD

pkg/kubernetes/kubernetes.go
```diff
@@ -203,6 +203,14 @@ func (c *ClientsCollection) Get(ctx context.Context, kind, namespace, name strin
 	return client.Get(ctx, name, options, subresources...)
 }
 
+func (c *ClientsCollection) Exists(ctx context.Context, kind, namespace, name string, options metav1.GetOptions, subresources ...string) bool {
+	_, err := c.Get(ctx, kind, namespace, name, options, subresources...)
+	if err != nil { // TODO: check for specific error
+		return false
```
mikkeloscar, line 209: It would make sense to return an error here (if it's not 404) such that it's clear why it failed and can be treated as a problem in cluster registry.

myaser: fixed

pkg/updatestrategy/aws_ec2.go
```diff
@@ -138,7 +138,7 @@ func (n *EC2NodePoolBackend) filterWithNodePool(nodePool *api.NodePool) []*ec2.F
 			},
 		},
 		{
-			Name: aws.String("tag:" + nodePoolTag),
+			Name: aws.String("tag:" + nodePoolTag), // TODO: this needs to be conditional
```
mikkeloscar, line 141: What do to about this TODO?

pkg/updatestrategy/aws_ec2.go
```diff
@@ -269,3 +273,93 @@ func (n *EC2NodePoolBackend) decommission(ctx context.Context, filters []*ec2.Fi
...
+func (r *KarpenterCRDNameResolver) NodeTemplateCRDName() string {
+	switch r.NodePoolCRDName {
+	case karpenterNodePoolResource:
+		return KarpenterEC2NodeClassResource
+	default:
+		return karpenterAWSNodeTemplateResource
+	}
+}
+
+func (r *KarpenterCRDNameResolver) referenceFieldName() string {
```
mikkeloscar, line 308: I don't see this used anywhere?

pkg/updatestrategy/aws_ec2.go
```diff
@@ -269,3 +273,93 @@ func (n *EC2NodePoolBackend) decommission(ctx context.Context, filters []*ec2.Fi
...
+	}
+}
+
+func (r *KarpenterCRDNameResolver) NodePoolConfigGetter(ctx context.Context, nodePool *api.NodePool) (*InstanceConfig, error) {
+	// CLM assumes that the node pool name is used for both the node-pool and the node-template that it references
+	NodeTemplate, err := r.k8sClients.Get(ctx, r.NodeTemplateCRDName(), "", nodePool.Name, v1.GetOptions{})
+	//if apierrors.IsNotFound(err) {
+	//	// the node pool have been deleted. thus the
+	//	return nil, nil
+	//}
```
mikkeloscar, line 348: nit: remove commented code

pkg/util/generics.go
```diff
@@ -8,3 +10,24 @@ func Contains[T comparable](s []T, e T) bool {
...
+			l.value, err = l.New()
+			l.New = nil
+		})
+	}
+	return l.value, err
+}
+
+func NewLazyOf[T any](newfunc func() (T, error)) *LazyOf[T] {
+	return &LazyOf[T]{New: newfunc}
+}
```
mikkeloscar, line 33: It's not completely clear to me why this lazy initialization is needed. Can't we just initialize when initializing the NodePool backend, or is it because the resources are not applied at that time?

The way to solve that would be to do the resolution whenever there is a need to lookup the CRD resources and not during the setup of the client. Since this is gonna be short lived I think we can keep it like this for now. But I would suggest to remove it long term to simplify the code.

myaser: yes, it is because the CRDs are not there during the initialization of the backend.

I tried to delay the initialization of the resolver but it was a bit difficult to pass all the parameters across the functions all the way to where it is needed. I thought this was "simpler" 😄 

> But I would suggest to remove it long term to simplify the code.

yea, sure, we can do this

pkg/updatestrategy/aws_ec2.go
myaser: 🤔 
actually, this may manage to mark old karpenter node-pools for decommission.
I will experiment with uncomment it and keep them to see if it makes a difference

mikkeloscar:
:+1:

myaser:
👍🏻

PR #749, opened 2024-02-21
Make decommissioning more graceful

mikkeloscar:
Sometimes you end up with a broken cluster where CLM can't read from the API. In those cases just fail more gracefully and continue decommissioning anyway.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #765, opened 2024-05-15
Add parameter to filter cloud provider.

gargravarr:
Adds a command line parameter configuring which cloud providers the Cluster Lifecycle Manager supports provisioning. The parameters has the default `zalando-aws`, currently the only provider supported.

Instead of not returning an error when the provisioner does not support a provider (my original idea), CLM uses the provider parameter to filter the cluster list coming from the Cluster Registry. In this way we can still segment provisioning by providers to different CLMs (when we support EKS for example).

Additionally I added some tests to increase test coverage.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -94,7 +94,7 @@ func (cfg *LifecycleManagerConfig) ParseFlags() string {
 	kingpin.Flag(
 		"provider",
 		"Cloud provider. Defaults to single provider \"zalando-aws\".",
-	).Default(defaultProvider).StringsVar(&cfg.Providers)
+	).Default(defaultProvider).EnumsVar(&cfg.Providers, "zalando-aws")
 	kingpin.Flag("config-source", "Config source specification (NAME:dir:PATH or NAME:git:URL). At least one is required.").StringsVar(&cfg.ConfigSources)
 	kingpin.Flag("directory", "Use a single directory as a config source (for local/development use)").StringVar(&cfg.Directory)
 	kingpin.Flag("concurrent-updates", "Number of updates allowed to run in parallel.").Default(defaultConcurrentUpdates).UintVar(&cfg.ConcurrentUpdates)
--- a/config/config.go
+++ b/config/config.go
@@ -93,7 +93,7 @@ func (cfg *LifecycleManagerConfig) ParseFlags() string {
 	kingpin.Flag("workdir", "Path to working directory used for storing channel configurations.").Default(defaultWorkdir).StringVar(&cfg.Workdir)
 	kingpin.Flag(
 		"provider",
-		"Cloud provider. Defaults to single provider \"zalando-aws\").",
+		"Cloud provider. Defaults to single provider \"zalando-aws\".",
 	).Default(defaultProvider).StringsVar(&cfg.Providers)
 	kingpin.Flag("config-source", "Config source specification (NAME:dir:PATH or NAME:git:URL). At least one is required.").StringsVar(&cfg.ConfigSources)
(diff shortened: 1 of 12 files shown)


config/config.go
```diff
@@ -89,6 +91,10 @@ func (cfg *LifecycleManagerConfig) ParseFlags() string {
 	kingpin.Flag("dry-run", "Don't make any changes, just print.").BoolVar(&cfg.DryRun)
 	kingpin.Flag("listen", "Address to listen at, e.g. :9090 or 0.0.0.0:9090").Default(defaultListener).StringVar(&cfg.Listen)
 	kingpin.Flag("workdir", "Path to working directory used for storing channel configurations.").Default(defaultWorkdir).StringVar(&cfg.Workdir)
+	kingpin.Flag(
+		"provider",
+		"Cloud provider. Defaults to single provider \"zalando-aws\".",
+	).Default(defaultProvider).StringsVar(&cfg.Providers)
```
mikkeloscar, line 97: nit: I think there is a way to define possible values for a flag via `Enum()` or so. Just to avoid setting an invalid value.

provisioner/provisioner.go
```diff
@@ -11,21 +11,30 @@ import (
...
+		DryRun          bool
+		ApplyOnly       bool
+		UpdateStrategy  config.UpdateStrategy
+		RemoveVolumes   bool
+		ManageEtcdStack bool
+	}
+)
+
+const (
+	ZalandoAWS ProviderID = "zalando-aws"
```
mikkeloscar, line 29: nit: Maybe name the variable `ZalandoAWSProvider` to make it more clear what it is?

config/config.go
gargravarr: Good idea, updated.

provisioner/provisioner.go
gargravarr: Renamed.

config/config.go
```diff
@@ -89,6 +91,10 @@ func (cfg *LifecycleManagerConfig) ParseFlags() string {
 	kingpin.Flag("dry-run", "Don't make any changes, just print.").BoolVar(&cfg.DryRun)
 	kingpin.Flag("listen", "Address to listen at, e.g. :9090 or 0.0.0.0:9090").Default(defaultListener).StringVar(&cfg.Listen)
 	kingpin.Flag("workdir", "Path to working directory used for storing channel configurations.").Default(defaultWorkdir).StringVar(&cfg.Workdir)
+	kingpin.Flag(
+		"provider",
+		"Cloud provider. Defaults to single provider \"zalando-aws\".",
+	).Default(defaultProvider).EnumsVar(&cfg.Providers, "zalando-aws")
```
AlexanderYastrebov, line 97: Maybe use `ZalandoAWSProvider` const here?

config/config.go
```diff
@@ -27,6 +27,7 @@ const (
 	defaultDrainForceEvictInterval          = "5m"
 	defaultDrainPollInterval                = "30s"
 	defaultUpdateStrategy                   = "clc"
+	defaultProvider                         = "zalando-aws"
```
AlexanderYastrebov, line 30: Maybe use `ZalandoAWSProvider` const here?

config/config.go
gargravarr: I tried at first, but got an annoying "import cycle not allowed" :slightly_frowning_face: , because `provisioner` package imports `config`.  

https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/de834d61afe158c963eecc128a581ba2bb320fb0/provisioner/clusterpy.go#L21

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #766, opened 2024-05-15
config: move provider id enum into config package

AlexanderYastrebov:
Move enum into config package to avoid import cycle and eliminate duplicate string literals.

Follow up on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/765/files#r1601499528

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -9,6 +9,14 @@ import (
 	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/updatestrategy"
 )
 
+// A provider ID is a string that identifies a cluster provider.
+type ProviderID string
+
+const (
+	// ZalandoAWS Provider is the provider ID for Zalando managed AWS clusters.
+	ZalandoAWSProvider ProviderID = "zalando-aws"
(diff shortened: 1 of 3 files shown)


gargravarr:
I was tempted to do this in my original PR, but didn't because it seems to me it makes more sense that this ID is in the `provisioner` package, where we implement provisioning for the supported providers, what do you think?

config/config.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/zalando-incubator/cluster-lifecycle-manager/pkg/updatestrategy"
 )
 
+// A provider ID is a string that identifies a cluster provider.
+type ProviderID string
+
+const (
+	// ZalandoAWS Provider is the provider ID for Zalando managed AWS clusters.
+	ZalandoAWSProvider ProviderID = "zalando-aws"
+)
```
gargravarr, line 18: nit: have a single `const` block.

AlexanderYastrebov:
I think its fine both ways so no need to change it right now.

PR #767, opened 2024-05-15
In-code kubectl apply

mikkeloscar:
Proof of Concept of porting the "kubectl apply" code from deployment-service to also utilize it in CLM. Basically this means making the `kubectl apply` operation via Go instead of calling the external `kubectl` binary.

The motivation for this change is that we currently restrict how many executions of `kubectl apply` we do concurrently as it's hard to control the memory. This has the side effect of limiting how many clusters we can process in parallel. The idea is that by having this in go we save memory by not having to run another process and we thus can do more in parallel.

The feature is behind a config-item `kubectl_apply_inline: true` which enables us to configure it only in a few clusters as a start to validate that it behaves as expected and gives the intended benefits.

## TODO

* [x] Roll out behind a config-item to test in selected clusters first
* [ ] ~Check if we still need to control concurrency.~


pkg/kubernetes/kubectl_apply.go
```diff
@@ -0,0 +1,207 @@
...
+			return err
+		}
+
+		// Create the resource if it doesn't exist
+		// First, update the annotation used by kubectl apply
+		if err := kubectlutil.CreateApplyAnnotation(obj, unstructured.UnstructuredJSONScheme); err != nil {
+			return err
+		}
+
+		// TODO: created log
```
demonCoder95, line 121: Do we want to resolve this? This looks more like a DEBUG log level entry, if we support a debug mode for CLM 🤔

pkg/kubernetes/kubectl_apply.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil
+	}
+
+	patcher := newPatcher(mapping, helper)
+	_, _, err = patcher.Patch(current, modified, manifest.Name, manifest.Namespace, manifest.Name, os.Stderr)
+	if err != nil {
+		return err
+	}
+
+	// TODO: maybe updated log?
```
demonCoder95, line 138: Same comment here as above.

demonCoder95:
> Check if we still need to control concurrency.

Does this require us to run any tests? I think it's fine to rollout this change and observe after a while, to check this concurrency requirement.

Otherwise, PR looks good 👍

pkg/kubernetes/kubectl_apply.go
mikkeloscar: Removed. I think it was from deployment-service code I copied.

mikkeloscar:
> > Check if we still need to control concurrency.
> 
> Does this require us to run any tests? I think it's fine to rollout this change and observe after a while, to check this concurrency requirement.

Yes, this we can best observe after rolling it out and enabling it in some clusters.

mikkeloscar:
:+1:

pkg/kubernetes/resource_manifest.go
```diff
@@ -0,0 +1,125 @@
...
+func (m *ResourceManifest) ToYaml() (string, error) {
+	result, err := yaml.Marshal(m.Content)
+	if err != nil {
+		return "", err
+	}
+	return string(result), nil
+}
+
+// FixupYamlValue converts the map types in a parsed opaque manifest
+// from map[interface{}]interface{}. Won't be needed once yaml.v3 is released.
```
tcondeixa: we are already using yaml.v3, or do we need a new version?

tcondeixa:
I think this is an important step to be faster to apply without memory constraints. What we would miss to port to Go from kubectl after this would be only the deletions?
This would require heavy testing and validations, we might end up implicitly changing some behaviours, but we need start moving to this approach.

tcondeixa:
This is behind a feature flag that we can validate for some time in some test cluster, so I think it's good to go

tcondeixa:
👍

pkg/kubernetes/resource_manifest.go
mikkeloscar: I have removed it, I think it doesn't matter anymore and was just left from the initial implementation.

Seems to work fine without :D

mikkeloscar:
> This is behind a feature flag that we can validate for some time in some test cluster, so I think it's good to go

I have been running this for pet clusters since forever so I'm quite confident that it works functionally. What I'm not confident about are the performance improvement.

I agree we can run it for some time in dev -> test and evaluate functionality and impact on performance.

mikkeloscar:
:+1:

tcondeixa:
👍

PR #768, opened 2024-05-17
Add EKS Provider ID.

gargravarr:
This Pull request adds the Provider ID for AWS EKS to the Cluster Lifecycle Manager (CLM). The `--provider` parameter will also accept `zalando-eks`. Implementing provisioning for EKS is left to another Pull Request, so the effect of configuring provider for `zalano-eks` will have the effect of CLM receiving an empty list of clusters from the Cluster Registry. This minimal setting will therefore allow to test and configure a deployment step of a CLM for Zalando EKS in a test cluster in our pipeline.


gargravarr:
:+

mikkeloscar:
:+1:

PR #769, opened 2024-05-22
Karpenter CLM rolling update and CLC support

myaser:
Allow CLM rolling update, and CLC update strategies to drain and delete karpenter nodes to maintain our SLAs

it is being tested here:
https://github.com/zalando-incubator/kubernetes-on-aws/pull/7560


pkg/updatestrategy/aws_ec2.go
```diff
@@ -203,7 +204,17 @@ func (n *EC2NodePoolBackend) Scale(context.Context, *api.NodePool, int) error {
...
+func (n *EC2NodePoolBackend) Terminate(ctx context.Context, pool *api.NodePool, node *Node, _ bool) error {
+	// terminating the instance using AWS api, it will also trigger karpenter interruption controller to
+	// delete the node and nodeClaim objects
+	instanceID := instanceIDFromProviderID(node.ProviderID, node.FailureDomain)
+	params := &ec2.TerminateInstancesInput{
+		InstanceIds: []*string{&instanceID},
+	}
+	_, err := n.ec2Client.TerminateInstancesWithContext(ctx, params)
+	if err != nil {
+		return fmt.Errorf("failed to terminate EC2 instances of the node pool '%s': %w", render.Render(pool), err)
```
mikkeloscar, line 216: Don't we have the node pool name on the `*Node` object? It seems a bit much to change the interface and pass `pool *api.NodePool` everywhere and then only use it in this error log :)

myaser: it is not used for this one only. the `ProfileNodePoolProvisioner` rely on the Profile to select the right backend (karpenter vs ASG). otherwise it defaults to ASG. it does not appear in the diff because ProfileNodePoolProvisioner was already used, since it was lacking the nodePool it was defaulting to ASG.

I can try to simplify the interface

pkg/updatestrategy/updatestrategy.go
```diff
@@ -108,6 +108,10 @@ func (n *ProfileNodePoolProvisioner) Scale(ctx context.Context, nodePool *api.No
...
 
 // Terminate terminates a node pool using the default provisioner.
-func (n *ProfileNodePoolProvisioner) Terminate(ctx context.Context, node *Node, decrementDesired bool) error {
-	return n.defaultProvisioner.Terminate(ctx, node, decrementDesired)
+func (n *ProfileNodePoolProvisioner) Terminate(ctx context.Context, nodePool *api.NodePool, node *Node, decrementDesired bool) error {
+	if provisioner, ok := n.profileMapping[nodePool.Profile]; ok {
+		return provisioner.Terminate(ctx, nodePool, node, decrementDesired)
+	}
+
+	return n.defaultProvisioner.Terminate(ctx, nodePool, node, decrementDesired)
```
myaser, line 116: @mikkeloscar 
all the methods of this `ProfileNodePoolProvisioner` rely on the node pool to decide the (karpenter vs ASG) logic. this was the only method that always defaults to ASG and was not supporting the split logic.
I thought passing the nodePool would be consistent with the interface of the rest of the methods

myaser: the ASG terminate logic relies on finding the ASG and decrement the number of desired nodes, ...etc and thus it will fail.

I can't modify the `ProfileNodePoolProvisioner` interface without `EC2NodePoolBackend` as they share the same interface

mikkeloscar: I see, I was the one who introduced this pattern :see_no_evil: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/316

Ideally we can refactor it later to use types instead of a profileMapping map... Doesn't make sense to refactor in this PR ofc.

provisioner/node_pools.go
```diff
@@ -237,14 +237,47 @@ func (p *KarpenterNodePoolProvisioner) Reconcile(ctx context.Context, _ updatest
...
+		if !inNodePoolList(&nodePool, karpenterPools) {
 			err := p.k8sClients.Delete(ctx, crdResolver.NodePoolCRDName, "", pr.GetName(), metav1.DeleteOptions{})
 			if err != nil {
 				return err
 			}
 		}
 	}
 
+	nodes, err := p.k8sClients.List(ctx, "nodes", "", metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("karpenter.sh/nodepool"),
```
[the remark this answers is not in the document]
mikkeloscar: Thanks for clarifying, makes sense!

provisioner/node_pools.go
```diff
@@ -237,14 +237,47 @@ func (p *KarpenterNodePoolProvisioner) Reconcile(ctx context.Context, _ updatest
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// also label nodes for decommissioning so CLC can speed up the decommissioning process,
+	// enforcing our SLAs since karpenter will wait indefinitely for the pod disruption budgets
+	// https://cloud.docs.zalando.net/reference/cluster-update-policy/#slas
```
mikkeloscar, line 258: nit: This depends on the updatestrategy used which for us is always CLC but could be the CLM one as well, so maybe we don't explicit mention CLC here?

Also we should not link to internal docs in a comment here.

myaser: updated

mikkeloscar:
:+1:

myaser:
👍🏻

PR #770, opened 2024-05-27
Switch to amazonlinux base image

mikkeloscar:
Switch to using Amazon Linux 2023 slim as base image to reduce the chance of CVEs.


linki:
What's missing here?

mikkeloscar:
@linki It should be fine now (rebased on master) I didn't move this forward initially because the build for ARM64 is quite slow, let's see how slow it is here now, it may be a blocker if it's too slow..

linki:
It's still very slow: 2 minutes for current master build vs. 8 minutes for this PR.

mikkeloscar:
> It's still very slow: 2 minutes for current master build vs. 8 minutes for this PR.

I don't understand how it can be so slow, it's just running `dnf` vs. `apk` `apk` is not slow at all on arm64 :shrug:

linki:
Still takes about twice the amount of time 🤷

linki:
👍

gargravarr:
:+1:

PR #776, opened 2024-06-03
provisioner: add list template function

AlexanderYastrebov:
This is useful for defining a fixed list instead of splitting a comma-separated string.

See also previous #711


mikkeloscar:
@AlexanderYastrebov Does this work if the list is defined one item per line? Could you add a test case for that if supported?

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #777, opened 2024-06-05
Add EKS support

gargravarr:
Adds support for provisioning an EKS cluster. Most of the features are described here https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/797

I added additional testing to some of the new components, like the `eksID` function and the `ProvisionModifier` interface. Unfortunately, code coverage still decreases because many of the changes are in sections using either the AWS or Kubernetes API and it would be a a big refactoring interface all of the APIs. I added, however, an initial interface to `awsAdapter` to properly test `ProvisionModifier`.


provisioner/zalando_eks.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, err
+	}
+	decodedCA, err := base64.StdEncoding.DecodeString(
+		clusterInfo.CertificateAuthority,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	res.APIServerURL = clusterInfo.Endpoint
```
linki, line 154: Currently, `api_server_url` for EKS clusters isn't the EKS endpoint, but rather the endpoint of the `zalando-iam-proxy`, i.e. for the current `teapot-eks` cluster it's https://kube-eks-1-zalando-iam-aws-proxy.teapot.zalan.do (`{{.Cluster.LocalID}}-zalando-iam-aws-proxy.{{ .Values.hosted_zone }}` from [here](https://github.com/zalando-incubator/kubernetes-on-aws/pull/7808/files#diff-86e8151404f9fb8ac4f9cc3ab78edfa6f1e9db7b45d4365e519adcffaafea331R12) and [here](https://github.bus.zalan.do/teapot/issues/issues/3667#issuecomment-5499773).)

If we run into issues then I think we should use this value here so that templates use the right endpoint.

gargravarr: I don't see the `APIServerURL` being used being used when CLM renders templates: https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/853861f3d4238c726901986751018471eddf3896/provisioner/template.go#L84-L138

So I guess it's fine.

gargravarr:
:+1:

provisioner/zalando_eks.go
gargravarr: That was obviously the wrong code section. But I searched for references of the `APIServerURL` field and didn't find any usage in the template rendering code.

linki:
👍

PR #779, opened 2024-06-18
Make provider support both self managed and EKS provisioning.

gargravarr:
Small hack to support provisioning EKS cluster via provider flag.


mikkeloscar:
:+1:

gargravarr:
:+1:

gargravarr:
Contributes to https://github.bus.zalan.do/teapot/issues/issues/3638

PR #780, opened 2024-06-18
controller: log cluster version

AlexanderYastrebov:
Log cluster version at processing start and finish.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -214,20 +214,25 @@ func (c *Controller) processCluster(updateCtx context.Context, workerNum uint, c
 	cluster := clusterInfo.Cluster
 	clusterLog := c.logger.WithField("cluster", cluster.Alias).WithField("worker", workerNum)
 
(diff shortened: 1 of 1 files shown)


controller/controller.go
```diff
@@ -214,20 +214,25 @@ func (c *Controller) processCluster(updateCtx context.Context, workerNum uint, c
...
 
-	clusterLog.Infof("Processing cluster (%s)", cluster.LifecycleStatus)
+	versionedLog := clusterLog
+	if clusterInfo.NextVersion != nil {
+		versionedLog = clusterLog.WithField("version", clusterInfo.NextVersion.String())
+	}
+
+	versionedLog.Infof("Processing cluster (%s)", cluster.LifecycleStatus)
 
 	err := c.doProcessCluster(updateCtx, clusterLog, clusterInfo)
```
AlexanderYastrebov, line 224: doProcessCluster uses clusterLog and therefore does not log cluster version to avoid clutter.

mikkeloscar:
:+1:

PR #789, opened 2024-07-03
Reduce dependabot frequency to weekly

mikkeloscar:
After #786 we get daily updates because of patch updates of e.g. aws-sdk-go: #788 , #787 

This seem like too much toil, so let's reduce the cadence of updates to weekly after we have verified the grouping of updates work.


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #792, opened 2024-07-08
Show config-item failing to decrypt

mikkeloscar:
Show which config-item is failing to decrypt to help debugging. Before the error just indicated that _some_ config-item decryption failed.


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #797, opened 2024-07-23
Implement EKS provisioning and support multiple provisioners.

gargravarr:
This Pull Request updates the `eks-support` branch with the following changes:

* Updates the CLM entry point support both Zalando AWS end EKS provisioning. I replaced the main provisioner instance with a map of provisioners. This map replaces the single `provisioner` parameter in the control loop, while command line picks the provisioner based on the specified provider.
* Splits EKS and self-managed provisioning in two `structs`: `ZalandoAWSProvisioner` and `ZalandoEKSProvisioner`:
    *  `ZalandoAWSProvisioner`: just calls the provisioning/decommissioning methods of `clusterPyProvisioner`.
    *  `ZalandoEKSProvisioner`: creates the EKS token source before calling provisioning/decommissioning methods of `clusterPyProvisioner`. Additionally for decommissioning passes the specific Certificate Authority data.
* Adds a `ProvisionModifier` interface. EKS Provisioning implements this interface, and the `clusterPyProvisioner` calls back `GetPostOptions`. When provisioning EKS, only after this stage it is possible to know some configuration necessary to provision a cluster, like: API Server endpoint, custom Certificate Authority, and Availability Zone information.
* Updates `clusterpyProvisioner` in the following:
    * Adds a `provider` field and implements a common `Supports` method.
    * Adds an optional `modifier`, which maybe used at a later provisioning stage 
    * Hides the interface from outside `provisioner` (the provisioners have a `clusterPyProvisioner` as a field).
    * Provides the AWS Adapter as a parameter to `provision()`. The EKS provisioner needs this adapter _before_ provisioning, to build a token source.
    * Refactors `prepareProvision()` to only apply default valuers to the various manifests. Gathering instance type information from AWS is now done by `provision`.
    * Calls back a modifier interface right after applying the first `cluster.yaml` CloudFormation.
    * Adds an optional Certificate Authority data to `decommission()`, necessary for decommissioning an EKS cluster.
* Updates the lower level Kubernetes functions (found in `pkg/kubernetes`) with support an optional parameter with custom Certificate Authority data. This certificate is necessary when provisioning an EKS cluster.
* Removes the token source parameter when creating the AWS Adapter. The token source is not used by the adapter.

This Pull Request does not add tests. A follow-up Pull Request to master will have more testing.

### DIFF
--- a/cmd/clm/main.go
+++ b/cmd/clm/main.go
@@ -115,39 +115,34 @@ func main() {
 
 	execManager := command.NewExecManager(cfg.ConcurrentExternalProcesses)
 
-	provisioners := map[provisioner.ProviderID]provisioner.Provisioner{}
-	for _, provider := range cfg.Providers {
-		switch provider {
-		case string(provisioner.ZalandoAWSProvider):
-			provisioners[provisioner.ZalandoAWSProvider] = provisioner.NewZalandoAWSProvisioner(
-				execManager,
-				clusterTokenSource,
-				secretDecrypter,
-				cfg.AssumedRole,
-				awsConfig,
-				&provisioner.Options{
-					DryRun:          cfg.DryRun,
-					ApplyOnly:       cfg.ApplyOnly,
-					UpdateStrategy:  cfg.UpdateStrategy,
-					RemoveVolumes:   cfg.RemoveVolumes,
-					ManageEtcdStack: cfg.ManageEtcdStack,
-				},
-			)
-		case string(provisioner.ZalandoEKSProvider):
-			provisioners[provisioner.ZalandoEKSProvider] = provisioner.NewZalandoEKSProvisioner(
-				execManager,
-				secretDecrypter,
-				cfg.AssumedRole,
-				awsConfig,
-				&provisioner.Options{
-					DryRun:         cfg.DryRun,
-					ApplyOnly:      cfg.ApplyOnly,
-					UpdateStrategy: cfg.UpdateStrategy,
-					RemoveVolumes:  cfg.RemoveVolumes,
-					Modifier:       &provisioner.ZalandoEKSModifier{},
-				},
-			)
-		}
+	provisioners := map[provisioner.ProviderID]provisioner.Provisioner{
+		provisioner.ZalandoAWSProvider: provisioner.NewZalandoAWSProvisioner(
+			execManager,
+			clusterTokenSource,
+			secretDecrypter,
+			cfg.AssumedRole,
+			awsConfig,
+			&provisioner.Options{
+				DryRun:          cfg.DryRun,
+				ApplyOnly:       cfg.ApplyOnly,
+				UpdateStrategy:  cfg.UpdateStrategy,
+				RemoveVolumes:   cfg.RemoveVolumes,
+				ManageEtcdStack: cfg.ManageEtcdStack,
+			},
+		),
+		provisioner.ZalandoEKSProvider: provisioner.NewZalandoEKSProvisioner(
+			execManager,
+			secretDecrypter,
+			cfg.AssumedRole,
+			awsConfig,
+			&provisioner.Options{
+				DryRun:         cfg.DryRun,
+				ApplyOnly:      cfg.ApplyOnly,
+				UpdateStrategy: cfg.UpdateStrategy,
+				RemoveVolumes:  cfg.RemoveVolumes,
+				Modifier:       &provisioner.ZalandoEKSModifier{},
+			},
+		),
 	}
 
 	configSource, err := setupConfigSource(execManager, cfg)
--- a/cmd/clm/main.go
+++ b/cmd/clm/main.go
@@ -115,13 +115,40 @@ func main() {
 
 	execManager := command.NewExecManager(cfg.ConcurrentExternalProcesses)
 
-	p := provisioner.NewClusterpyProvisioner(execManager, clusterTokenSource, secretDecrypter, cfg.AssumedRole, awsConfig, &provisioner.Options{
-		DryRun:          cfg.DryRun,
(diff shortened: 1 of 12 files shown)


provisioner/clusterpy_test.go
```diff
@@ -533,7 +533,7 @@ func TestDecommissionDoesNotSupportProvider(t *testing.T) {
 		Provider: "zalando-eks",
 	}
 
-	p := clusterpyProvisioner{}
+	p := ZalandoEKSProvisioner{}
```
linki, line 536: Why is the expected result `ErrProviderNotSupported`? Shouldn't the `ZalandoEKSProvisioner` support the cluster when `cluster.Provider` is `zalando-eks`?

provisioner/clusterpy_test.go
```diff
@@ -533,7 +533,7 @@ func TestDecommissionDoesNotSupportProvider(t *testing.T) {
...
 	err := p.Decommission(context.TODO(), nil, cluster)
 	assert.Equal(t, ErrProviderNotSupported, err)
```
linki, line 538: Why is the expected result `ErrProviderNotSupported`? Shouldn't the `ZalandoEKSProvisioner` support the cluster when `cluster.Provider` is `zalando-eks`?

cmd/clm/main.go
```diff
@@ -115,13 +115,40 @@ func main() {
...
 
-	p := provisioner.NewClusterpyProvisioner(execManager, clusterTokenSource, secretDecrypter, cfg.AssumedRole, awsConfig, &provisioner.Options{
-		DryRun:          cfg.DryRun,
-		ApplyOnly:       cfg.ApplyOnly,
-		UpdateStrategy:  cfg.UpdateStrategy,
-		RemoveVolumes:   cfg.RemoveVolumes,
-		ManageEtcdStack: cfg.ManageEtcdStack,
-	})
+	provisioners := map[provisioner.ProviderID]provisioner.Provisioner{}
+	for _, provider := range cfg.Providers {
```
linki, line 119: nit: I think we can consider avoiding the dynamic provider list going forward. The flag `--provider` merely controls which provider is initialized and added to the struct here. The selection of the right provider for each cluster is still being done by the `.Provider` field on the cluster struct.

In other words, what prevents us from unconditionally doing the following?
```go
provisioners[provisioner.ZalandoAWSProvider] = provisioner.NewZalandoAWSProvisioner(...)
provisioners[provisioner.ZalandoEKSProvider] = provisioner.NewZalandoEKSProvisioner(...)
```

pkg/kubernetes/kubernetes.go
```diff
@@ -405,35 +420,42 @@ func (k *KubeCTLRunner) KubectlExecute(ctx context.Context, args []string, stdin
...
-		fmt.Sprintf("--certificate-authority=%s", tmpfile.Name()), // TODO: eks
-	}, args...)
+	args = append(kubeCtlOpts, args...)
 	if stdin != "" {
 		args = append(args, "-f", "-")
 	}
 
 	newCommand := func() *exec.Cmd {
-		cmd := exec.Command(args[0], args[1:]...)
+		cmd := exec.Command("kubectl", args[1:]...)
```
linki, line 458: Can you double check that this is correct? With your changes above, `args[0]` is the first _flag_ instead of the binary name like before, hence this hard-coded value here. But then the next parameter should be `args[0:]...` instead of `args[1:]...`.

provisioner/zalando_eks.go
```diff
@@ -0,0 +1,178 @@
...
+	options *Options,
+) Provisioner {
+	provisioner := &ZalandoEKSProvisioner{
+		clusterpyProvisioner: clusterpyProvisioner{
+			provider:          ZalandoEKSProvider,
+			awsConfig:         awsConfig,
+			assumedRole:       assumedRole,
+			execManager:       execManager,
+			secretDecrypter:   secretDecrypter,
+			manageMasterNodes: false,
```
linki, line 41: Is `manageMasterNodes` actually used anywhere? I can't find it in this PR and since it is first introduced here I would expect it here somewhere.

provisioner/clusterpy_test.go
```diff
@@ -524,7 +524,7 @@ func TestProvisionDoesNotSupportProvider(t *testing.T) {
 		Provider: "zalando-eks",
 	}
 
-	p := clusterpyProvisioner{}
+	p := ZalandoEKSProvisioner{}
 	err := p.Provision(context.TODO(), nil, cluster, nil)
 	assert.Equal(t, ErrProviderNotSupported, err)
```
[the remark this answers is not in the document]
gargravarr: Many things were wrong here: the test as you observed, but also the `Supports` method, which with the intent of saving some code lines I put the implementation in `clusterpyProvisioner`. Without the `provider` initialized the test of course was passing.

I fixed the test, removed the `provider` field and put the implementation of `Supports()` in the actual provisioners.

cmd/clm/main.go
gargravarr: I agree, no need to loop and switch. Updated.

provisioner/zalando_eks.go
gargravarr: :facepalm: forgot to uncomment code, update here https://github.com/zalando-incubator/cluster-lifecycle-manager/commit/cf74c3c29154fb74f67ef8079f95454b1035b50a

pkg/kubernetes/kubernetes.go
gargravarr: You are right. args[0] now is the first flag, I tested it in the [Go playground](https://go.dev/play/p/IiNWAyxl8ds). Will update.

linki:
👍

gargravarr:
:+1:

PR #801, opened 2024-08-06
Update Cluster Registry entry when provisioning an EKS cluster.

gargravarr:
This Pull Request updates a new EKS cluster entry in the Cluster Registry with config items for the EKS API Server URL and Certificate Authority Data. These items are needed by other components to connect to the EKS Cluster.


provisioner/zalando_eks.go
```diff
@@ -133,6 +142,21 @@ func (z *ZalandoEKSProvisioner) Decommission(
...
+	return &ZalandoEKSModifier{
+		clusterRegistry: clusterRegistry,
+	}
+}
+
+// GetPostOptions returns the configuration only known after deploying the first
+// CloudFormation stack.
+//
+// This includes the API server URL, the Certificate Authority data, and the
+// subnets. Additionally GetPostOptions update the cluster
```
linki, line 159: > Additionally GetPostOptions update the cluster

This comment looks unfinished.

Can we do the config item update outside of this function somewhere? When reading the `GetPostOptions` function name one doesn't expect it to have any side effects (updating the cluster registry).

linki: Can we do the config item update outside of this function somewhere? When reading the `GetPostOptions` function name one doesn't expect it to have any side effects (updating the cluster registry).

registry/http_test.go
```diff
@@ -0,0 +1,177 @@
...
+			continue
+		}
+
+		err := registry.UpdateConfigItems(
+			&api.Cluster{ID: testClusterID, ConfigItems: tc.configItems},
+		)
+
+		require.NoError(t, err)
+		for key := range tc.configItems {
+			require.Contains(t, tc.configItems, key)
```
linki, line 100: This might be superflous because it's always true since we're iterating over the keys already: `for key := range tc.configItems`

provisioner/zalando_eks.go
gargravarr: How about we rename the interface method to something like `UpdateProvisioning`? To describe better what the method actually does.

registry/http_test.go
gargravarr: This is obviously not what I want to test. I wanted to check that the config items in `clusterInRegistry` contain all keys. Fixed.

provisioner/zalando_eks.go
linki: Fine. But then I would use the opportunity to rename it further to a more familiar style. How if we change the following to this:

image

* Changing `Modifier` to `PostCreationHook`
* Changing `GetPostOptions` to `Execute` or `Call` (e.g. `PostCreationHook.Call()`) which _can_ have side effects and returns `postOptions` (which we should then rename to `hookResult` or something like that)

gargravarr: > Changing Modifier to PostCreationHook

My only issue with this name is that we are calling the method not _after_ creating the cluster, but _during_ cluster creation.

linki: Since it's just a rename we can also do this in another PR and allow us some time to come up with the proper naming.

linki:
👍

provisioner/zalando_eks.go
gargravarr: I updated the interface to `CreationHook`. In reality a provisioner can choose when the hook applies.

linki:
👍

gargravarr:
:+1:

PR #802, opened 2024-08-07
Only update EKS config items when needed.

gargravarr:
The config item update implemented in https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/801 used the wrong API call from Cluster Registry and updated all clulster's config items, including keys with defaults.

This Pull Requests ensures that CLM will update the EKS config items only if the keys do not exist, or if the corresponding values are different.


linki:
👍

gargravarr:
:+1:

PR #804, opened 2024-08-08
Add OIDC issuer url as config item to the cluster

linki:
For [ServiceAccount for IAM roles](https://github.bus.zalan.do/teapot/cluster-registry-deploy/blob/master/deploy/prod/iam-role.yaml#L22-L34) to work, we need to configure the right OIDC provider. As a first step, make the URL available via Config Items.

The ConfigItem can then be used to configure the right parameters here: https://github.com/zalando-incubator/kubernetes-on-aws/commit/f16824fcea0f74e25a92959142d4472342c7aef4

```yaml
  oidc-provider-arn: "arn:aws:iam::{{accountID .Cluster.InfrastructureAccount}}:oidc-provider/oidc.eks.eu-central-1.amazonaws.com/id/EKSID"
  oidc-subject-key: "oidc.eks.eu-central-1.amazonaws.com/id/EKSID:sub"
```

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -123,9 +123,9 @@ type (
 
 	// awsInterface is an interface containing methods of an AWS Adapter.
 	//
-	// ProvisionModifier uses this interface in the GetPostOptions method.
+	// CreationHook uses this interface in the Execute method.
 	awsInterface interface {
-		GetEKSClusterCA(cluster *api.Cluster) (*EKSClusterInfo, error)
+		GetEKSClusterDetails(cluster *api.Cluster) (*EKSClusterDetails, error)
 	}
 )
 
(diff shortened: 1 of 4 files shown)


linki:
👍

provisioner/aws.go
```diff
@@ -771,6 +771,7 @@ func tagsToMap(tags []*ec2.Tag) map[string]string {
 type EKSClusterInfo struct {
 	Endpoint             string
 	CertificateAuthority string
+	OIDCIssuerURL        string
 }
 
 func (a *awsAdapter) GetEKSClusterCA(cluster *api.Cluster) (*EKSClusterInfo, error) {
```
gargravarr, line 777: We should rename this method to something like `GetEKSClusterDetails`. The method return more than the CA Data. I suggest the suffix "Details" because that's where this data shows is in the Overview of the cluster in the AWS Console.

gargravarr:
:+1:

linki:
👍

PR #808, opened 2024-09-01
migrate aws package ec2 client to aws-sdk-go-v2

demonCoder95:
This migrates parts of the code to `aws-sdk-go-v2`. This is used in other tooling as it has exported functions so migrating it helps migrate those tools as well.


mikkeloscar:
This was done in #897

PR #816, opened 2024-09-17
Store the account name as a field on the cluster object

linki:
Similar to `Owner`, fetches the account details from the `/infrastructure-accounts` endpoint and stores the account name (e.g. eks-playground) under the `.AccountName` field so it can be used in templating.


linki:
👍

demonCoder95:
👍

PR #817, opened 2024-09-17
Use unique cluster alias as identifier rather than redundant eksID

linki:
Using `eksID()` generates redundant and unreadable EKS cluster names, as well as other resources that are based on it.

For example, a cluster with alias `playground-2` will have an EKS resource named `aws--123456789012--eu-central-1--kube-1` for which the ARN will become a long, unreadable and redundant:

```
arn:aws:eks:eu-central-1:123456789012:cluster/aws--123456789012--eu-central-1--kube-1
```

[Update] Since the merge of https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/818 the ARN will look like this:

```
arn:aws:eks:eu-central-1:123456789012:cluster/kube-1
```

This isn't too bad, but it's unclear which cluster "kube-1" is. One has to use the AWS account ID to figure that out. Subsequent clusters in the same account would likely use "kube-2", "kube-3" etc. which isn't really helpful.

With this change the cluster will show in the UI as `playground-2` and will always be unique as long as we pick unique cluster aliases. The ARN will become more useful and correct:

```
arn:aws:eks:eu-central-1:123456789012:cluster/playground-2
```

- [ ] Let's decommission any incompatible EKS cluster before merging this.

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -777,7 +777,7 @@ type EKSClusterDetails struct {
 
 func (a *awsAdapter) GetEKSClusterDetails(cluster *api.Cluster) (*EKSClusterDetails, error) {
 	resp, err := a.eksClient.DescribeCluster(&eks.DescribeClusterInput{
-		Name: aws.String(eksID(cluster.ID)),
+		Name: aws.String(cluster.Alias),
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


demonCoder95:
👍

demonCoder95:
LGTM, once we decommission `playground-eks` we can merge I guess.

linki:
It's a little to much change at the moment. I'll put it on hold.

PR #821, opened 2024-09-27
Fix eks decommissioning for pet clusters

mikkeloscar:
During decommission of EKS clusters CLM tried to get EKS endpoint URL and EKS endpoint CA from the config-item in cluster-registry. For pet clusters those config-items won't be set and thus the decommissioning won't work correctly.

Instead, simply discover the values dynamically from the EKS cluster.


mikkeloscar:
:+1:

linki:
👍

PR #822, opened 2024-09-30
Add ipv6 eks support

mikkeloscar:
This adds some missing part for getting eks based ipv6 clusters working.

1. Change the AZInfo structure to also include IPv6 CIDR information such that it can be passed to templates via the `.Value.subnet_ipv6_cidrs` varaible. This is so we can pass the list of ipv6 subnets to skipper-ingress for determining internal vs. external traffic.
2. Expose the EKS IPv6 Service CIDR as config-item `service_ipv6_cidr` and add a template function `addressNFromIPV6CIDR` which given a ipv6 cidr and a number `n` can return the nth address from the CIDR. This is done to have a predictable way to get a service IP for the skipper-internal service which is used for `ingress.cluster.local` hostnames.
3. Clean up some CF Output/EKSSubnet handling code, which was left over from some past experiments but are no longer used in the eks branch.

### DIFF
--- a/provisioner/aws_az.go
+++ b/provisioner/aws_az.go
@@ -4,14 +4,20 @@ import (
 	"sort"
 )
 
+// SubnetInfo has information about a subnet.
+type SubnetInfo struct {
+	SubnetID        string
+	SubnetIPV6CIDRs []string
+}
+
 // AZInfo tracks information about available AZs based on explicit restrictions or available subnets
 type AZInfo struct {
-	subnets map[string]string
+	subnets map[string]SubnetInfo
 }
 
 // RestrictAZs returns a new AZInfo that is restricted to provided AZs
 func (info *AZInfo) RestrictAZs(availableAZs []string) *AZInfo {
-	result := make(map[string]string)
+	result := make(map[string]SubnetInfo)
 
 	for _, az := range availableAZs {
 		if subnet, ok := info.subnets[az]; ok {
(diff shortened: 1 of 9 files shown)


provisioner/aws_az_test.go
```diff
@@ -8,10 +8,16 @@ import (
...
+		subnets: map[string]SubnetInfo{
+			"eu-central-1a": SubnetInfo{
+				SubnetID: "subnet-1a",
+			},
+			"eu-central-1b": SubnetInfo{
+				SubnetID: "subnet-1b",
+			},
+			"eu-central-1c": SubnetInfo{
+				SubnetID: "subnet-1c",
+			},
```
demonCoder95, line 20: Do we plan to expand this to include tests for `v6` as well?

provisioner/clusterpy.go
```diff
@@ -312,11 +313,8 @@ func (p *clusterpyProvisioner) provision(
 		if postOptions.APIServerURL != "" {
 			cluster.APIServerURL = postOptions.APIServerURL
 		}
-		if postOptions.AZInfo != nil {
-			azInfo = postOptions.AZInfo
-		}
-		for k, v := range postOptions.TemplateValues {
-			values[k] = v
```
demonCoder95, line 319: Is this the code cleanup mentioned in the PR description?

provisioner/template.go
```diff
@@ -581,6 +582,26 @@ func nodeCIDRMaxPods(maskSize int64, extraCapacity int64) (int64, error) {
...
+	if ip == nil {
+		return "", fmt.Errorf("invalid IP address: %s", ipNet.IP)
+	}
+	ip = ip.Mask(ipNet.Mask)
+	for i := 0; i < n; i++ {
+		ip[15]++
+	}
+	return ip.String(), nil
+}
+
```
demonCoder95, line 604: We should definitely cover this with a unit test

provisioner/aws_az.go
```diff
@@ -4,14 +4,20 @@ import (
 	"sort"
 )
 
+// SubnetInfo has information about a subnet.
+type SubnetInfo struct {
+	SubnetID        string
+	SubnetIPV6CIDRs []string
```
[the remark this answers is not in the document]
mikkeloscar: A subnet can have multiple ipv6 cidrs associated. It's unlikely that we will have more than one, but this is making it "forwards" compatible.

provisioner/clusterpy.go
demonCoder95: Okay, I just reread (3) from the description and it turns out it is. Then I suppose it's fine to remove it and replace it like it's done here.

provisioner/aws_az.go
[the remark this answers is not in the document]
demonCoder95: Makes sense. Thanks!

provisioner/template.go
mikkeloscar: done

provisioner/aws_az_test.go
mikkeloscar: Added a test for the new method introduced here as well

mikkeloscar:
:+1:

PR #824, opened 2024-10-04
Support dynamic IPv4 EKS service CIDR

mikkeloscar:
Follow up to #822 

Also support providing EKS IPv4 service CIDRs to templates as they are also dynamic by default like for ipv6.

I change the logic a bit to be more generic with a general `service_cidr` config-item which can have an ipv4 or ipv6 CIDR depending on the type set for the cluster. The template function `nthAddressFromCIDR` (renamed from `AddressNFromIPv6CIDR`) has been made generic to handle both ipv4 and ipv6. This makes the templating logic a bit simpler where we need this.

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -116,7 +116,7 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		},
 		"nodeCIDRMaxNodesPodCIDR":               nodeCIDRMaxNodes,
 		"nodeCIDRMaxPods":                       nodeCIDRMaxPods,
-		"addressNFromIPv6CIDR":                  addressNFromIPv6CIDR,
+		"nthAddressFromCIDR":                    nthAddressFromCIDR,
 		"parseInt64":                            parseInt64,
 		"generateJWKSDocument":                  generateJWKSDocument,
 		"generateOIDCDiscoveryDocument":         generateOIDCDiscoveryDocument,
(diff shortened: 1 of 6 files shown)


katyanna:
:+1:

mikkeloscar:
:+1:

provisioner/template.go
```diff
@@ -582,22 +582,33 @@ func nodeCIDRMaxPods(maskSize int64, extraCapacity int64) (int64, error) {
...
+		if ip == nil {
+			return "", fmt.Errorf("invalid IP address: %s", ipNet.IP)
+		}
+		nthByte = 15
 	}
+
 	ip = ip.Mask(ipNet.Mask)
 	for i := 0; i < n; i++ {
-		ip[15]++
+		ip[nthByte]++
```
demonCoder95, line 611: @mikkeloscar , looking at this a bit closer I see a few issues. 

For example, if I give the CIDR `192.168.0.0/30`. The current implementation will provide me the `n=10` address as `192.168.0.10` (incrementing the nth byte by `10`) which is incorrect. Since the subnet only really has 4 addresses (192.168.0.0-3). 

I think we need to catch the case where n is potentially > number of host addresses in a subnet so we don't end up in this situation.

(We may never end up in it, but it just came to my mind when looking at the code again..)

AlexanderYastrebov: I think this could be solved like:
```go
package main

import (
	"fmt"
	"log"
	"math/big"
	"net"
)

func main() {
	const (
		cidr = "2a05:d014:9c0:bf05::/64"
		n    = 50
	)

	_, ipNet, err := net.ParseCIDR(cidr)
	if err != nil {
		log.Fatal(err)
	}

	bigN := big.NewInt(int64(n))
	ones, bits := ipNet.Mask.Size()

	if bigN.BitLen() > bits-ones {
		log.Fatalf("n %d requires nBits %d which is greater than 2^%d", n, bigN.BitLen(), bits-ones)
	}

	intIP := new(big.Int)
	intIP.SetBytes(ipNet.IP)
	intIP.Add(intIP, bigN)
	addr := net.IP(intIP.Bytes())

	fmt.Printf("addr: %s\n", addr)
}
```

mikkeloscar: Thanks both!

It definitely makes sense to change this to make it more robust. What I did just resolved what we needed for now, but I'm all for an implementation without edge cases in case we need it for something else in the future.

PR #825, opened 2024-10-07
Expose vpc ipv6 cidrs as Values

mikkeloscar:
This adds the optional `.Values.vpc_ipv6_cidrs` field to the template context such that the vpc IPv6 CIDRs can be used e.g. in skipper-ingress config as needed for: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8304

it also adds a standard `join` template function which makes it easier to work with `[]string`

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -95,6 +95,7 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"azID":                 azID,
 		"azCount":              azCount,
 		"split":                split,
+		"join":                 strings.Join,
 		"eksID":                eksID,
 		"mountUnitName":        mountUnitName,
 		"accountID":            accountID,
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -95,7 +95,7 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"azID":                 azID,
 		"azCount":              azCount,
 		"split":                split,
-		"join":                 strings.Join,
+		"join":                 sprig.GenericFuncMap()["join"],
 		"eksID":                eksID,
 		"mountUnitName":        mountUnitName,
 		"accountID":            accountID,
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -252,6 +252,11 @@ func (p *clusterpyProvisioner) provision(
 		return err
 	}
 
+	vpcIPv6CIDRs := make([]string, 0, len(vpc.Ipv6CidrBlockAssociationSet))
+	for _, ipv6Cidr := range vpc.Ipv6CidrBlockAssociationSet {
+		vpcIPv6CIDRs = append(vpcIPv6CIDRs, aws.StringValue(ipv6Cidr.Ipv6CidrBlock))
+	}
+
 	values := map[string]interface{}{
 		subnetsValueKey:             azInfo.SubnetsByAZ(),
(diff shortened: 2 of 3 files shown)


provisioner/template.go
```diff
@@ -95,6 +95,7 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"azID":                 azID,
 		"azCount":              azCount,
 		"split":                split,
+		"join":                 strings.Join,
```
AlexanderYastrebov, line 98: Maybe it makes sense to revert argument order like here https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/621/files#diff-807fefe6626b30fd095cee0c1c6cf81f2219f513b244dd13ba403395786734fdR281-R285 to enable piped join like `{{ $foo | join "," }}`

AlexanderYastrebov:
:+1:

provisioner/template.go
mikkeloscar: Changed!

mikkeloscar:
:+1:

provisioner/template.go
```diff
@@ -91,10 +91,14 @@ func renderTemplate(context *templateContext, file string) (string, error) {
...
-		"azCount":              azCount,
-		"split":                split,
+		"asgSize": asgSize,
+		"azID":    azID,
+		"azCount": azCount,
+		"split":   split,
+		"join":    join,
+		"stringSlice": func(args ...string) []string {
+			return args
+		},
```
AlexanderYastrebov, line 101: I think we can use existing list for the purpose, can't we?

mikkeloscar: The problem is `list` produces `[]interface{}` and `join` takes `[]string`. Making `join` take `[]interface{}` seem ugly.

AlexanderYastrebov: > Making join take []interface{} seem ugly.

sprig does that though https://github.com/Masterminds/sprig/blob/fc7fc0d6a0377bca7049c4a99e80b85f222d8caf/strings.go#L199-L201

Maybe we can just use it (like we use `indent`) - it allowes piped input and takes case of typo conversion...

AlexanderYastrebov: Another option is to add `toStrings` helper
https://github.com/Masterminds/sprig/blob/fc7fc0d6a0377bca7049c4a99e80b85f222d8caf/functions.go#L186
```
{{ list "foo" "bar" | toStrings | join "," }}
```

AlexanderYastrebov: > stringSlice

If we go this way then lets name it `stringList` or `slist` to make it aligned with existing `list`.

mikkeloscar: Ok, then let's just do the sprig `join` function. I have changed it to use that one.

provisioner/template.go
```diff
@@ -257,6 +258,12 @@ func split(s string, d string) []string {
 	return strings.Split(s, d)
 }
 
+// join is a template function that returns a single string of slice elements joined with a separator.
+func join(sep string, elems []string) string {
+	// arguments in reverse order to enable piped calls
+	return strings.Join(elems, sep)
+}
```
AlexanderYastrebov, line 265: This we dont need anymore

mikkeloscar: Ups, fixed!

mikkeloscar:
:+1:

PR #827, opened 2024-10-25
Support separate subnets for LBs and workers

mikkeloscar:
Add support for choosing different subnets for running Load Balancers and nodes. Until now they all used the same subnets leading to challenges when wanting to use different ingress and egress rules.

The subnets are selected based on tags:

* LB: `kubernetes.io/role/elb`
* Nodes: `kubernetes.io/role/node`

For safe roll out, we will first add the `kubernetes.io/role/node` tag to existing subnets with the tag: `kubernetes.io/role/elb` such that there is no change for running clusters as a change will require a migration.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -221,18 +221,18 @@ func (p *clusterpyProvisioner) provision(
 
 	// find the best subnet for each AZ
 	azInfoLBs := selectSubnetIDs(subnets, subnetELBRoleTagName)
-	azInfoWorkers := selectSubnetIDs(subnets, subnetNodeRoleTagName)
+	azInfoNodes := selectSubnetIDs(subnets, subnetNodeRoleTagName)
 
 	// if availability zones are defined, filter the subnet list
 	if azNames, ok := cluster.ConfigItems[availabilityZonesConfigItemKey]; ok {
 		azInfoLBs = azInfoLBs.RestrictAZs(strings.Split(azNames, ","))
-		azInfoWorkers = azInfoWorkers.RestrictAZs(strings.Split(azNames, ","))
+		azInfoNodes = azInfoNodes.RestrictAZs(strings.Split(azNames, ","))
 	}
 
 	// optional config item to hard-code subnets. This is useful for
 	// migrating between subnets.
 	if _, ok := cluster.ConfigItems[subnetsConfigItemKey]; !ok {
-		cluster.ConfigItems[subnetsConfigItemKey] = azInfoWorkers.SubnetsByAZ()[subnetAllAZName]
+		cluster.ConfigItems[subnetsConfigItemKey] = azInfoNodes.SubnetsByAZ()[subnetAllAZName]
 	}
 
 	apiURL, err := url.Parse(cluster.APIServerURL)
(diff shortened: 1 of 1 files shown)


provisioner/clusterpy.go
```diff
@@ -219,16 +220,18 @@ func (p *clusterpyProvisioner) provision(
...
+	azInfoWorkers := selectSubnetIDs(subnets, subnetWorkerRoleTagName)
 
 	// if availability zones are defined, filter the subnet list
 	if azNames, ok := cluster.ConfigItems[availabilityZonesConfigItemKey]; ok {
-		azInfo = azInfo.RestrictAZs(strings.Split(azNames, ","))
+		azInfoLBs = azInfoLBs.RestrictAZs(strings.Split(azNames, ","))
+		azInfoWorkers = azInfoWorkers.RestrictAZs(strings.Split(azNames, ","))
 	}
 
 	// TODO legacy, remove once we switch to Values in all clusters
```
AlexanderYastrebov, line 232: > // TODO legacy, remove once we switch to Values in all clusters

Does this comment somehow change meaning or is it maybe outdated?

mikkeloscar: True, this is being used for controlled migrations between subnets. Let me update the comment as well.

provisioner/clusterpy.go
```diff
@@ -258,9 +262,10 @@ func (p *clusterpyProvisioner) provision(
 	}
 
 	values := map[string]interface{}{
-		subnetsValueKey:             azInfo.SubnetsByAZ(),
-		availabilityZonesValueKey:   azInfo.AvailabilityZones(),
-		subnetIPV6CIDRsKey:          strings.Join(azInfo.SubnetIPv6CIDRs(), ","),
+		subnetsValueKey:             azInfoWorkers.SubnetsByAZ(),
+		availabilityZonesValueKey:   azInfoWorkers.AvailabilityZones(),
+		subnetIPV6CIDRsKey:          strings.Join(azInfoWorkers.SubnetIPv6CIDRs(), ","),
+		"lb_subnets":                azInfoLBs.SubnetsByAZ(),
```
AlexanderYastrebov, line 268: This is new, do we need in a similar change in provisionNodePool ?

AlexanderYastrebov: I guess this is handled by `azInfo: azInfoWorkers,`

provisioner/clusterpy.go
```diff
@@ -219,16 +220,19 @@ func (p *clusterpyProvisioner) provision(
 	}
 
 	// find the best subnet for each AZ
-	azInfo := selectSubnetIDs(subnets)
+	azInfoLBs := selectSubnetIDs(subnets, subnetELBRoleTagName)
+	azInfoWorkers := selectSubnetIDs(subnets, subnetNodeRoleTagName)
```
AlexanderYastrebov, line 224: Here we use "node role" and "workers", the annotation is also `kubernetes.io/role/node` and not `kubernetes.io/role/worker` - should this be consistent and use either "node" or "worker"?

AlexanderYastrebov:
:+1:

provisioner/clusterpy.go
mikkeloscar: Updated the variable to be consistent with the tag name after that was changed.

AlexanderYastrebov:
:+1:

provisioner/clusterpy.go
mikkeloscar: This will be used like this: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8440

mikkeloscar:
This change require that we set `kubernetes.io/role/node` tag on existing subnets to be backwards compatible.

I have tested what happens if we should have forgotten to set the tags in an existing cluster. The CF stack fails with:

image


And rolls back without disruption. So there is no impact if it should be the case.

mikkeloscar:
:+1:

PR #829, opened 2024-10-31
Add a cluster method Name to get canonical clusterName based on provider

mikkeloscar:
This adds a cluster method `Name()` which returns the canonical cluster ID/Name based on the cluster Provider.

This is helpful as for non-EKS we use the full `ID` as cluster ID e.g. via tags on resources and for EKS we use only `LocalID` as the `ID` is not compatible with EKS because of `:` in the value.

This makes it simpler to template such that both EKS and non-EKS is supported.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -287,7 +287,7 @@ func (p *clusterpyProvisioner) provision(
 	}
 
 	// create or update the etcd stack
-	if p.manageEtcdStack && cluster.Provider == string(ZalandoAWSProvider) {
+	if p.manageEtcdStack && cluster.Provider == api.ZalandoAWSProvider {
 		etcdKMSKeyARN, err := awsAdapter.resolveKeyID(etcdKMSKeyAlias)
 		if err != nil {
 			return err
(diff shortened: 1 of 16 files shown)


provisioner/template.go
```diff
@@ -839,3 +840,9 @@ func scaleQuantity(quantityStr string, factor float32) (string, error) {
 	}
 	return quantity.String(), nil
 }
+
+// clusterName returns the canonical cluster name based on the cluster
+// provider.
+func clusterName(cluster api.Cluster) (string, error) {
```
linki, line 846: This doesn't need to return an error.

linki: I assume we can't use `cluster.Name()` in the templates directly?

provisioner/clusterpy.go
```diff
@@ -1017,8 +1017,13 @@ func (p *clusterpyProvisioner) downscaleDeployments(
...
 
 func (p *clusterpyProvisioner) listClusterStacks(_ context.Context, adapter *awsAdapter, cluster *api.Cluster) ([]*cloudformation.Stack, error) {
+	clusterName, err := clusterName(*cluster)
+	if err != nil {
+		return nil, err
+	}
+
 	includeTags := map[string]string{
-		tagNameKubernetesClusterPrefix + cluster.ID: resourceLifecycleOwned,
+		tagNameKubernetesClusterPrefix + clusterName: resourceLifecycleOwned,
```
linki, line 1026: `cluster.Name()` should work here, too.

provisioner/template.go
mikkeloscar: Dropped this function as we can use `.Cluster.Name` in templates. The test is updated to use this.

provisioner/clusterpy.go
mikkeloscar: Fixed

mikkeloscar:
:+1:

linki:
👍

PR #830, opened 2024-11-05
Optionally support dedicated pod subnets

mikkeloscar:
ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8510

Adds support for discovering subnets for pods if using [Custom Networking](https://docs.aws.amazon.com/eks/latest/best-practices/custom-networking.html)


AlexanderYastrebov:
:+1:

PR #838, opened 2025-01-06
Fix generating EKS token

mikkeloscar:
This fixes getting a correct token for EKS clusters such that CLM can authenticate to an EKS cluster cross account.

This broke in #834 as the aws-iam-authenticator library was updated and the call was changed to not pass the existing session.

The problem was that when not passing the session, it falls back to setup a session from the current environment. When CLM is running centrally the current environment is an IAM role in the central account instead of an assumed role in the target account which has access to the targeted EKS cluster. The fix is to get a token in a way where we use the assumed role session to setup an STS client. Other API calls were removed from the upstream library in https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/750

### DIFF
--- a/pkg/aws/eks/token.go
+++ b/pkg/aws/eks/token.go
@@ -2,6 +2,7 @@ package eks
 
 import (
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sts"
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #840, opened 2025-01-13
Add support for internel node subnets

mikkeloscar:
This adds support for finding subnets tagged with `kubernetes.io/role/internal-node` and expose them a `.Values.internal_node_subnets` in templates.

The purpose of this is to be able to discover and launch a subnet of node pools in internal subnets. This is useful for special instances running with NAT gateways or simply for running nodes in internal subnets.

A concrete use case is getting p5.48xlarge instances working with EFA. Those can't automatically get a public IP so we need another way to give them internet access e.g. via NAT.


AlexanderYastrebov:
:+1:

mikkeloscar:
:+1:

PR #841, opened 2025-01-17
Add append template function

mikkeloscar:
Add `append` template functions which can append an item to a list. It's intentially only implemented for `string` as we don't need more than this for now.

It's needed for https://github.com/zalando-incubator/kubernetes-on-aws/pull/8786

### DIFF
--- a/provisioner/template.go
+++ b/provisioner/template.go
@@ -131,6 +131,7 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"indent":                                sprig.GenericFuncMap()["indent"],
 		"dict":                                  dict,
 		"list":                                  list,
+		"append":                                strAppend,
 		"scaleQuantity":                         scaleQuantity,
 		"instanceTypeCPUQuantity": func(instanceType string) (string, error) {
 			return instanceTypeCPUQuantity(context, instanceType)
@@ -281,6 +282,11 @@ func dict(args ...interface{}) (map[string]interface{}, error) {
 
 func list(args ...interface{}) []interface{} {
 	return args
+}
+
+func strAppend(list []string, item string, moreItems ...string) []string {
+	items := append(list, item)
+	return append(items, moreItems...)
 }
 
 func eksID(id string) string {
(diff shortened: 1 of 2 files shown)


provisioner/template.go
```diff
@@ -283,6 +284,10 @@ func list(args ...interface{}) []interface{} {
 	return args
 }
 
+func strAppend(list []string, item string) []string {
```
AlexanderYastrebov, line 287: Does it make sense to have vararg items like:
```go
func strAppend(list []string, item string, other ...string) []string
```
?

Not sure if that works but we can have a test :)

provisioner/template.go
```diff
@@ -283,6 +284,10 @@ func list(args ...interface{}) []interface{} {
...
+	return append(list, item)
```
AlexanderYastrebov, line 288: Here we may want to clone the list first to not accidentally modify it.

mikkeloscar: Doesn't `append` make a copy?

https://go.dev/play/p/_EM10XVQVn-

mikkeloscar: Doesn't `append` make a copy?

https://go.dev/play/p/J8LQXgGgBxB

provisioner/template.go
mikkeloscar: Updated!

mikkeloscar:
:+1:

provisioner/template.go
AlexanderYastrebov: Only when there is no capacity left

provisioner/template.go
AlexanderYastrebov: I suggested to have item+other to prohibit `{{ append .Values.data.items }}` which makes no sense.

In go stdlib they missed that for min/max functions, see https://github.com/golang/go/issues/66179#issuecomment-1989029221

provisioner/template.go
AlexanderYastrebov: OTOH it does exactly what append does so maybe it OK.

provisioner/template.go
mikkeloscar: Makes sense, updated!

AlexanderYastrebov:
:+1:

PR #848, opened 2025-02-19
Delete leaked ENIs on decommissioning

mikkeloscar:
In EKS clusters there can be ENIs left over from the AWS VPC CNI which blocks the deletion of the cluster stack during decommissioning because the ENI refer to a security group in the stack.

To work around this issue which is mostly present during e2e cluster cleanup, this PR adds a hook to find and delete such ENIs before the cluster stack is being deleted.


linki:
👍

PR #852, opened 2025-03-24
Fix git test by using standard branch name

mikkeloscar:
For me, locally, the git related tests were always failing with a strange 128 exit code.

```
--- FAIL: TestGitGet (0.07s)
    git_test.go:55:
        	Error Trace:	/home/moscar/projects/teapot/cluster-lifecycle-manager/channel/git_test.go:55
        	            				/home/moscar/projects/teapot/cluster-lifecycle-manager/channel/git_test.go:80
        	Error:      	Received unexpected error:
        	            	exit status 128
        	Test:       	TestGitGet
FAIL
```

The reason turned out be that I have configured: `defaultBranch = main` in my local config, but the tests assumes `master` is the default branch.

This PR changes the tests to do `git init -b main` and use main in all cases. This way it will be predictable no matter what the local config is. And we get rid of the use of `master`.

This also adds stderr to the git command err such that it's easier to debug such issues in the future as opposed to only see the exit code.

### DIFF
--- a/channel/git.go
+++ b/channel/git.go
@@ -1,6 +1,7 @@
 package channel
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

PR #854, opened 2025-03-25
Improve decommissioning of karpenter nodes

mikkeloscar:
This improves the logic for decommissioning karpenter nodes during cluster decommissioning to avoid leaking karpenter nodes which can't be cleaned up and block full cluster decommissioning.

The problem today is that decommissioning depends on Kubernetes API access during decommissioning. If  a cluster is broken, e.g. in an e2e scenario, then CLM can fail to clean up karpenter nodes.

Instead of relying on the Kubernetes API, which is only really used to get an instance tag during decommissioning, instead only rely on AWS API and delete any EC2 instance matching the node pool and cluster tag.

Additionally, this PR removes the logic for handling legacy karpenter CRD names (provisioner and AWS template). We only need to handle the new resource names (nodepool and EC2NodeClass) so we can drop the complexity of supporting both.

### DIFF
--- a/pkg/updatestrategy/aws_ec2.go
+++ b/pkg/updatestrategy/aws_ec2.go
@@ -24,13 +24,10 @@ import (
 )
 
 const (
-	karpenterProvisionerTag          = "karpenter.sh/provisioner-name"
-	karpenterNodePoolTag             = "karpenter.sh/nodepool"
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

PR #858, opened 2025-04-22
Remove legacy karpenter_pools_enabled logic

mikkeloscar:
Remove legacy logic for determining order of apply steps based on `karpenter_pools_enabled=true` config-item.

The config-item was removed in https://github.com/zalando-incubator/kubernetes-on-aws/pull/9147 making CLM behave in the old way even though karpenter is the default in all clusters.

This currently breaks e2e and node pools are not created before manifests are applied.

Fix it by just dropping the custom logic that anyway was always true in practice.

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -421,25 +421,21 @@ func (p *clusterpyProvisioner) provision(
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #866, opened 2025-06-20
Add member function to cluster holding the trust relationship

linki:
Alternative implementation to https://github.com/zalando-build/deployment-service/pull/268.

Allows to change [deployment-config](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/deployment-service/01-config.yaml) to the following:

```yaml
  worker-role-arn: "{{.Cluster.WorkerRoleARN}}"
  oidc-provider-arn: "{{.Cluster.OIDCProviderARN}}"
  oidc-subject-key: "{{.Cluster.OIDCSubjectKey}}"
  iam-role-trust-relationship-template: '{{.Cluster.IAMRoleTrustRelationshipTemplate}}'
```

The first three variables hold the trust relationship information for the current cluster. They are backwards compatible. The last variable holds a combined trust relationship of all clusters in the account (according to cluster registry).


api/cluster.go
```diff
@@ -210,3 +226,87 @@ func (cluster Cluster) Name() string {
...
+}
+
+func (cluster Cluster) OIDCProvider() string {
+	if cluster.Provider == ZalandoEKSProvider {
+		return strings.TrimPrefix(cluster.ConfigItems["eks_oidc_issuer_url"], "https://")
+	}
+
+	hostedZone, err := util.GetHostedZone(cluster.APIServerURL)
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
linki: Changed it to store the values upfront (with error handling) and then it's only accessed during templating.

mikkeloscar:
:+1:

pkg/util/hosted_zone.go
```diff
@@ -1,21 +1,21 @@
...
 	"fmt"
 	"net/url"
 	"strings"
 )
 
-// getHostedZone gets derrive hosted zone from an APIServerURL.
-func getHostedZone(APIServerURL string) (string, error) {
-	url, err := url.Parse(APIServerURL)
+// GetHostedZone gets the hosted zone from an APIServerURL.
+func GetHostedZone(apiServerURL string) (string, error) {
```
AlexanderYastrebov, line 10: Since we promote this into util and make exported maybe it makes sense to pass cluster object (or make it a method of cluster) - this way in the future we may use e.g. dedicated cluster field or config item to store hosted zone.

linki: Good idea! But I would like to keep it out of this PR, hoping no one relies on the function outside of the repo. I was merely moving it around to avoid duplication. It's all quite hacky in first place 😕

mikkeloscar:
:+1:

PR #874, opened 2025-08-14
export etcd instance information to the rendering function

myaser:
export etcd instance information to the rendering function

### DIFF
--- a/provisioner/clusterpy.go
+++ b/provisioner/clusterpy.go
@@ -487,6 +487,7 @@ func createOrUpdateEtcdStack(
 		nodePool:      nil,
 		instanceTypes: instanceTypes,
 	}
+	values["instance_types_info"] = instanceTypes
 
 	s3Path, err := renderer.RenderAndUploadFiles(values, bucketName, etcdKmsKeyARN)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


provisioner/clusterpy.go
```diff
@@ -496,6 +496,8 @@ func createOrUpdateEtcdStack(
 	logger.Debugf("Uploaded generated files to %s", s3Path)
 	values[s3GeneratedFilesPathValuesKey] = s3Path
 
+	values["instance_type_info"] = instanceTypes
```
zaklawrencea, line 499: Why not keep the key name `etcd_instance_type_info` for clarity? There is already a key for the `values` map: `instanceInfoKey` (== `"InstanceInfo"`), but this one does not refer to etcd instances.

linki, line 499: Keep an eye out for the rendering function on line 491: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/874/files#diff-b49199d424770fa971be5fcea8c0799afe324a9aa259c7083b2271c383721333R491 Looks like it renders there, too.

zaklawrencea: Thanks for clarifying. On the naming, now that I understand the purpose better, I don't really have a much better suggestion :/

myaser:
👍🏻

PR #880, opened 2025-09-01
Use cluster alias for canonical name for EKS clusters

linki:
All EKS clusters (except some old playground clusters) use `.Alias == .LocalID`, so we can use the alias for the canonical name.

If we decide to switch back to `.Alias != .LocalID` for EKS, then this change will result in a better Name for the EKS cluster resource in AWS and it will also be backwards-compatible with existing clusters.


tcondeixa:
I think we might want to close this without merge depending on the outcome of the current task

linki:
No, we can merge it. If we continue with he current approach, then `Alias == LocalID` and this is a no-op. If we pick anything else for `LocalID` going forward (during cluster creation), we'll want to use `Alias` for the `Name` instead of `LocalID`.

provisioner/template.go
```diff
@@ -96,7 +96,6 @@ func renderTemplate(context *templateContext, file string) (string, error) {
 		"azCount":              azCount,
 		"split":                split,
 		"join":                 sprig.GenericFuncMap()["join"],
-		"eksID":                eksID,
```
linki, line 99: I didn't find any use of `eksID()` in kubernetes-on-aws. I think we might be able to drop it fyi @mikkeloscar

linki:
👍

PR #914, opened 2026-03-19
chore(deps): update k8s.io/* to v0.35.2

ni9hty:
This PR updates k8s.io/* to v0.35.2 and go to 1.25

- exclude SA1019: manager.Uploader is deprecated: superceded by feature/s3/transfermanager lint check.
https://github.com/aws/aws-sdk-go-v2/discussions/3306 (since 30.01.2026)

As it needs several code changes

PR #921, opened 2026-05-05
Switch to s3 transfermanager

mikkeloscar:
Migrate away from the legacy s3 manager interface to the new transfermanager interface.

ref: #914


tcondeixa:
👍

mikkeloscar:
:+1:

PR #924, opened 2026-05-07
Add very basic custom CF stack support

mikkeloscar:
This adds very basic support for custom CF stacks to enable deploying resources outside of the "main" manifest repository that need some AWS resources like IAM roles or S3 buckets.

### How it works:

The manifest repo can now define a new folder `cluster/cf-manifests` which is a flat folder with an optional list of CloudFormation Manifests in the same format as for e.g. the cluster.yaml stack:

```yaml
Metadata:
  StackName: "custom-stack"
  Tags:
    application: "custom-application"

Resources:
...
```

The manifests defined in that folder are applied _after_ the cluster.yaml stack, but before any Kubernetes manifests (from `cluster/manifests` folder). The manifests are applied in lexicographical order, there is _no_ support for dependencies between the stacks.

Multiple config-sources can define stacks like this, they are just merged into a slice and applied one by one.

On cluster decommissioning the stacks are cleaned up in the reverse order e.g. before the `cluster.yaml` stack. The cleanup is out of the box because the stacks will be tagged with the cluster ID.

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -2,6 +2,7 @@ package provisioner
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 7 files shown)


tcondeixa:
Do we have all the existing feature for templating as we have for kubernetes manifests and the cluster CF stack?

can we delete the stacks at the end in the reverse `lexicographical` order? at least we have a consistent way to ensure creation and deletion.

As a safeguard I would check if there are no stack with the same stack name before start the process, so we fail fast and avoid potential issues. I assume this is not that difficult to do, otherwise we can do it in a separated PR later.

mikkeloscar:
> As a safeguard I would check if there are no stack with the same stack name before start the process, so we fail fast and avoid potential issues. I assume this is not that difficult to do, otherwise we can do it in a separated PR later.

We can do but you can argue it's the same with Kubernetes manifests today. I'm not opposed to make it more safe, but we also need to align if we want to invest this much in CLM in the longer term.

> can we delete the stacks at the end in the reverse lexicographical order? at least we have a consistent way to ensure creation and deletion.

In theory, but we need to store the file names on the stacks to know the order when we delete. To me it seems like a low priority thing it only matters if we start to make stacks depend on each other in some ways which we should not need to do for now.

The purpose of the feature is to unblock existing use cases, namely ACID manifests from CDP to CLM, it's not about adding more advanced CF features that we can use.

provisioner/aws.go
```diff
@@ -199,17 +200,54 @@ func tagMapToCloudformationTags(tags map[string]string) []cftypes.Tag {
...
+type CFManifest struct {
+	Metadata struct {
+		StackName string            `yaml:"StackName"`
+		Tags      map[string]string `yaml:"Tags"`
+	} `yaml:"Metadata"`
+	Template string
+	Path     string
+}
+
+func (a *awsAdapter) applyCFManifest(ctx context.Context, cfManifest *CFManifest, cluster *api.Cluster, s3BucketName string) error {
```
tcondeixa: nit: we can pass only the cluster.Name() as string

tcondeixa:
> In theory, but we need to store the file names on the stacks to know the order when we delete. To me it seems like a low priority thing it only matters if we start to make stacks depend on each other in some ways which we should not need to do for now.
> 
> The purpose of the feature is to unblock existing use cases, namely ACID manifests from CDP to CLM, it's not about adding more advanced CF features that we can use.

I think it should be fine for ACID use-cases, IIRC the dependencies they have are already part of the same stack (e.g. bucket and bucket policy), so it will be solved by Cloudformation itself.

I don't see any issue with proceeding with this PR, and it will allow us to move on with the migration from CDP to CLM to central teams (observability can also use this). We can do some tests adding a dummy CF stack and trying to change it, the changes should not affect the behaviour unless we add CF stacks.

provisioner/aws.go
mikkeloscar: Updated it.

tcondeixa:
@mikkeloscar let's merge this one?

tcondeixa:
👍

mikkeloscar:
:+1:

PR #925, opened 2026-05-07
cmd: add `render` command

MustafaSaber:
Introduce a new command to render files against CCM config-item file.

This is different from `dry-run` that requires access to AWS & real k8s cluster api server. The goal from this render command is to have a quick easy way to render templates from `kubernetes-on-aws` against set of config-items

I used locally like this see
https://github.com/zalando-incubator/kubernetes-on-aws/pull/11194 description

```bash
~ ./build/clm render --directory="./../kubernetes-on-aws" --configs-file="./../ccm/clusters/teapot.yaml"
```

Can be used later in CCM to get templates before & after a change and post git diff in github comment.

This PR can also be isolated in another repo/binary, but if we go this way we will need to export some functions


tcondeixa:
IMO it would be better we render both the templates and values using the real sources (e.g cluster registry), having a read-only role should be enough to have exactly what will be applied. We have these features in AILM if you want to check https://github.com/zalando-build/aws-infrastructure-lifecycle-manager/tree/master/cmd

MustafaSaber:
> IMO it would be better we render both the templates and values using the real sources (e.g cluster registry), having a read-only role should be enough to have exactly what will be applied. We have these features in AILM if you want to check https://github.com/zalando-build/aws-infrastructure-lifecycle-manager/tree/master/cmd

So the idea itself is appealing but we should work on the values source?

The idea about no AWS values is that if we want to integrate elsewhere (ccm for example) this will require some creds with readonly in all clusters (which I don't know how much hustle is it)

tcondeixa:
> > IMO it would be better we render both the templates and values using the real sources (e.g cluster registry), having a read-only role should be enough to have exactly what will be applied. We have these features in AILM if you want to check https://github.com/zalando-build/aws-infrastructure-lifecycle-manager/tree/master/cmd
> 
> So the idea itself is appealing but we should work on the values source?

I'm not very familiar with this codebase, but I would expect we have the building blocks to create a command.
IMO rendering the templates has higher priority, so the values can be done in a later PR.

> 
> The idea about no AWS values is that if we want to integrate elsewhere (ccm for example) this will require some creds with readonly in all clusters (which I don't know how much hustle is it)

I need to check the extend of the utilisation of the AWS values on the fly, but at least for cluster-registry it should be easy to use, both running locally or in kubernetes with a zalando token.

tcondeixa:
@MustafaSaber from what I can see the CLM only uses values from the defaults plus the cluster registry to get the necessary values to render the templates, there is no need to have access to AWS or Kubernetes

MustafaSaber:
> @MustafaSaber from what I can see the CLM only uses values from the defaults plus the cluster registry to get the necessary values to render the templates, there is no need to have access to AWS or Kubernetes

Will check, maybe I missed it

mikkeloscar:
> @MustafaSaber from what I can see the CLM only uses values from the defaults plus the cluster registry to get the necessary values to render the templates, there is no need to have access to AWS or Kubernetes

It does get some dynamic values like VPC and subnet information currently: https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/22f6a91d47f2ad0294078a196037a29a89330ef9/provisioner/clusterpy.go#L186-L201

tcondeixa:
> > @MustafaSaber from what I can see the CLM only uses values from the defaults plus the cluster registry to get the necessary values to render the templates, there is no need to have access to AWS or Kubernetes
> 
> It does get some dynamic values like VPC and subnet information currently:
> 
> https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/22f6a91d47f2ad0294078a196037a29a89330ef9/provisioner/clusterpy.go#L186-L201

I did not find these ones, thanks. Is this only about the VPC ID and subnets? IMO it might be a good moment to add these by default in the config-items when creating the clusters. IMO having this dynamic values might create issues if something changes in AWS, and it's more difficult to reason about since nothing changes in the repo nor the cluster registry.

MustafaSaber:
> IMO it might be a good moment to add these by default in the config-items when creating the clusters. IMO having this dynamic values might create issues if something changes in AWS, and it's more difficult to reason about since nothing changes in the repo nor the cluster registry.

@diegoroccia 's change is reading them if they already exists https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/c738839daa9942ef262b703c61aa4d760435cee2/provisioner/clusterpy.go#L848 so  may pick this up from his PR but I think it should be a smaller separate PR.

szuecs:
I tested it and I get this:
```
% ./clm render --directory="../../zaland-incubator/kubernetes-on-aws" --configs-file="../../zalando-build/teapot-cluster-config-manager/clusters/teapot.yaml"
=== Cluster: stub ===

FATA[0000] Failed to load values file: failed to unmarshal file "../../zalando-build/teapot-cluster-config-manager/clusters/teapot.yaml": yaml: unmarshal errors:
  line 26: cannot unmarshal !!seq into string
```

I added this to understand the error better:
```diff
% git diff
diff --git cmd/clm/main.go cmd/clm/main.go
index 39f1ad2..723d050 100644
--- cmd/clm/main.go
+++ cmd/clm/main.go
@@ -95,11 +95,11 @@ type configsFile struct {
 func loadConfigsFile(path string) (map[string]string, error) {
        data, err := os.ReadFile(path)
        if err != nil {
-               return nil, err
+               return nil, fmt.Errorf("failed to read file %q: %w", path, err)
        }
        var vf configsFile
        if err := yaml.Unmarshal(data, &vf); err != nil {
-               return nil, err
+               return nil, fmt.Errorf("failed to unmarshal file %q: %w", path, err)
        }
        return vf.ConfigItems, nil
 }
```

MustafaSaber:
> I tested it and I get this:
> 
> ```
> % ./clm render --directory="../../zaland-incubator/kubernetes-on-aws" --configs-file="../../zalando-build/teapot-cluster-config-manager/clusters/teapot.yaml"
> === Cluster: stub ===
> 
> FATA[0000] Failed to load values file: failed to unmarshal file "../../zalando-build/teapot-cluster-config-manager/clusters/teapot.yaml": yaml: unmarshal errors:
>   line 26: cannot unmarshal !!seq into string
> ```


I think because `teapot.yaml` have

```yaml
 key:
    - namespace: default
      service_account: sa2
    - namespace: kube-system
      service_account: sa1
```

which I didn't accommodate for

MustafaSaber:
> I think because `teapot.yaml` have
> 
> ```yaml
>  key:
>     - namespace: default
>       service_account: sa2
>     - namespace: kube-system
>       service_account: sa1
> ```
> 
> which I didn't accommodate for

To fix this, we can duplicate code from CCM to read this or reuse CCM code. It makes sense to reuse the code but in this case it's better to just have this command as a standalone repo and just use both repos code to read schema and apply

PR #926, opened 2026-05-08
Add SiblingCluster method

mikkeloscar:
This adds a method to the *Cluster type called `SiblingCluster` which returns nil or the SiblingCluster. Sibling cluster is defined as a ready cluster in the same region and account. If an account have more than 2 clusters in the same region, the oldest _other_ cluster is defined as the Sibling.

The purpose of this feature is to allow cross references in templates for migration purpose. A concrete example is postgres migration from non-EKS to EKS which has the following bucket policy requirement.

```yaml
{{- if ne .Cluster.SiblingCluster() nil }}
PostgreS3BucketEKSPolicy:
  Type: AWS::S3::BucketPolicy
  Properties:
    Bucket: !Ref PostgreS3Bucket
    PolicyDocument:
      Version: 2012-10-17
      Statement:
        - Effect: "Allow"
          Principal:
            AWS: "arn:aws:iam::{{{AWS_ACCOUNT_ID}}}:role/postgres-pod-role-{{ .Cluster.SiblingCluster().LocalID }}"
          Action:
            - "s3:ListBucket"
          Resource:
            - "arn:aws:s3:::{{{WAL_S3_BUCKET}}}"
        - Effect: "Allow"
          Principal:
            AWS: "arn:aws:iam::{{{AWS_ACCOUNT_ID}}}:role/postgres-pod-role-{{ .Cluster.SiblingCluster().LocalID }}"
          Action:
            - "s3:GetObject"
          Resource:
            - "arn:aws:s3:::{{{WAL_S3_BUCKET}}}/*"
{{- end }}
```

### DIFF
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -318,6 +318,29 @@ func policyStatements(workerRole string, identityProvider string, subjectKey str
 	}
 }
 
+// SiblingCluster returns the sibling cluster of the given cluster, or nil if
+// none exists. A sibling cluster is defined as another ready cluster in the
+// same account and region. If there are more than two clusters in the account
+// and region, the oldest sibling is returned.
+func (cluster Cluster) SiblingCluster() *Cluster {
+	var sibling *Cluster
+	for _, c := range cluster.AccountClusters {
+		if c.ID == cluster.ID {
+			continue
+		}
+		if c.Region != cluster.Region {
+			continue
+		}
+		if c.LifecycleStatus != models.ClusterLifecycleStatusReady {
+			continue
+		}
+		if sibling == nil || c.CreatedAt.Before(sibling.CreatedAt) {
+			sibling = c
+		}
+	}
+	return sibling
+}
+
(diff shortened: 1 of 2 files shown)


tcondeixa:
Not sure if providing the oldest is better than just return the first occurrence, but it's not an issue anyway and this should not happen.

tcondeixa:
👍

tcondeixa:
JFYI @linki since I remember you were checking also this topic.

mikkeloscar:
> Not sure if providing the oldest is better than just return the first occurrence, but it's not an issue anyway and this should not happen.

In the other method it's assumed that `AccountClusters` is sorted by creation time, so there it's implicit. Here it just checks to not make the assumption of the order of the list.

But it's at least important that it's by age or some other predictable order, we don't want that a new cluster suddenly changes the order, and thereby which is the sibling, e.g. because it was ordered by name or so.

mikkeloscar:
:+1:

PR #930, opened 2026-05-21
add module to the cf logging to be easy to filter by repo

tcondeixa:
add module to the cf logging to be easy to filter by repo and log success operation

`extractModuleFromPath` was used instead of a bigger refactor the list of CF manifests, but it might be a better option in the future.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #938, opened 2026-06-23
add simple cf-deletions similar to the k8s resources deletions

tcondeixa:
add simple cf-deletions similar to the k8s resources deletions.

Using the same strategy with `pre_apply` and `post_apply`, and pass a list of objects with the Cloudformation stack names. This allow us to later extend it without breaking changes, as mentioned in the comment https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/938#discussion_r3466278293

```yaml
pre_apply: # everything defined under here will be deleted before applying the manifests
- stackName: stack-name-1
- stackName: stack-name-2

post_apply: # everything defined under here will be deleted after applying the manifests
- stackName: stack-name-1
- stackName: stack-name-2
```

Other important decision:
- if the file does not exists in one/all config repos nothing will be deleted and it will not cause an error
- deletion of CF stacks is done after the deletion of the kubernetes resources
- using a different file name `cf-deletions.yaml` to avoid confusion (I'm not 100% sure about this decison), and the file is inside `cf-manifests` folder.
- the file `cf-deletions.yaml` is excluded from the CF apply since the files for CF manifests are not nested.

Test:
- add cf-deletions.yaml from some sources and not from others
- having sources without the file defined
- add a non existent stack to the cf-deletions.yaml

More details on the current use-case here: https://github.com/zalando-build/dbaas-kubernetes-manifests/pull/146

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,13 +182,15 @@ The `cf-deletions.yaml` file inside the `cf-manifests` folder has the following 
 
 ```yaml
 pre_apply: # everything defined under here will be deleted before applying the manifests
-- stack-name-1
-- stack-name-2
+- stackName: stack-name-1
+- stackName: stack-name-2
 
 post_apply: # everything defined under here will be deleted after applying the manifests
-- stack-name-1
-- stack-name-2
+- stackName: stack-name-1
+- stackName: stack-name-2
 ```
+
+Each stack deletion is specified as an object with a `stackName` field. This structured format allows for future extensions, such as tag-based selection or other selection criteria.
 
 
 ## Configuration defaults
--- a/README.md
+++ b/README.md
@@ -122,7 +122,12 @@ clusters:
 
 By default the Cluster Lifecycle Manager will just apply any manifest defined
 in the manifests folder. In order to support deletion of deprecated resources
-the CLM will read a `deletions.yaml` file of the following format:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -171,6 +176,21 @@ It is possible to specify deletion options via optional:
 - `propagation_policy` - one of "Orphan", "Background" or "Foreground" - corresponds to `kubectl delete --cascade` flag
 - `grace_period_seconds` - corresponds to `kubectl delete --grace-period` flag
 
+### Cloudformation deletions
+
+The `cf-deletions.yaml` file inside the `cf-manifests` folder has the following format:
+
+```yaml
+pre_apply: # everything defined under here will be deleted before applying the manifests
+- stack-name-1
```
mikkeloscar: Maybe we could use a format like: `- stackName: <name>`, this would allow extending this to e.g. tag based selection or similar if we should need it, easier if type is not just a string (stackname).

tcondeixa: make sense, it's a minor change that would allow us to improve without breaking changes.
I changed it in the latest commit and also adjusted the description of the PR, please check again @mikkeloscar

tcondeixa:
👍

mikkeloscar:
:+1:

PR #941, opened 2026-06-29
try to cleanup all possible cases to avoid disk size always growing

tcondeixa:
try to cleanup all possible cases to avoid disk size always growing

The introduction of the `git gc --prune=now` might introduce a few extra seconds per git update, but it will be sure that all references not needed are deleted right away.


mikkeloscar:
:+1:

tcondeixa:
👍

PR #944, opened 2026-07-10
separate function to wait for deletion that check expected end states

tcondeixa:
separate function to wait for deletion that check expected end states specific for deletion

This was detected when decommission a cluster where the exported values were used by other stacks as imports.
The stack was not deleted, and Cloudformation returned `UPDATE_COMPLETE` that was a valid end state.
The cluster progressed to `decommissioned` but the CF stack remained there, so I had to remove the dependencies and manually delete it after.

### DIFF
--- a/provisioner/aws.go
+++ b/provisioner/aws.go
@@ -55,6 +55,7 @@ var (
 	errUpdateRollbackFailed   = fmt.Errorf("wait for stack failed with %s", cftypes.StackStatusUpdateRollbackFailed)
 	errDeleteFailed           = fmt.Errorf("wait for stack failed with %s", cftypes.StackStatusDeleteFailed)
 	errTimeoutExceeded        = fmt.Errorf("wait for stack timeout exceeded")
+	errDeleteCanceled         = fmt.Errorf("stack deletion was canceled")
 )
 
 type (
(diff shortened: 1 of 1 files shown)


provisioner/aws.go
```diff
@@ -404,6 +405,42 @@ func (a *awsAdapter) waitForStack(ctx context.Context, waitTime time.Duration, s
...
+		if err != nil {
+			return err
+		}
+
+		switch stack.StackStatus {
+		case cftypes.StackStatusDeleteComplete:
+			return nil
+		case cftypes.StackStatusDeleteFailed:
+			err = errDeleteFailed
+		case cftypes.StackStatusUpdateComplete,
```
mikkeloscar, line 420: Will the status change on the initial delete call or is there a risk that this waiting checks and get the original stackset before the deletion has "started"?

tcondeixa: they don't ensure (or it's not clear to me) it starts before we receive the response, so it's better to cover this case.

"Once the call completes successfully, stack deletion starts."

mikkeloscar: Ok, then the risk is that this checks "wrongly" fails and then when CLM retries it would have gotten into the delete status, unless there is the problem with stack dependencies as you outlined.

mikkeloscar:
:+1:

provisioner/aws.go
tcondeixa: I tried this approach, but not sure if the added complexity worth it (and assumes the clocks are relatively in-sync). can you share your thoughts @mikkeloscar 
https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/944/changes/31a0ee0200c520f4ee72e8543523ad62f27a0bf8

mikkeloscar:
:+1:

tcondeixa:
👍

PR #949, opened 2026-07-14
Use Karpenter types instead of unstructured

mikkeloscar:
This changes calls to Karpenter related resources from using a dynamic client returning `*unstructured.Unstructured` to use a controller client which allows working with the native Karpenter Go structs.

The purpose is to simplify working with the types and not having to unwrap unstructured maps.


mikkeloscar:
:+1:

zaklawrencea (approved):

mikkeloscar:
:+1:

PR #950, opened 2026-07-14
Use Karpenter drift signals for outdated nodes

mikkeloscar:
Historically CLM has used UserData, AMI ID and tags to define whether a node was outdated from the desired configuration and needed to be rotated.

With the introduction of EKS AMIs: https://github.com/zalando-incubator/kubernetes-on-aws/pull/11643 this no longer works as the UserData is dynamic for those AMIs. E.g. the UserData defined on the EC2NodeClass is just a subset of the final userdata set on the instance, so it's no longer possible to compare the userdata to determine if a node is outdated.

Karpenter provides the Drift feature and indicates on a NodeClaim if it's drifted from the desired.

This PR changes the logic to use the Drift signal of NodeClaims instead of trying to determine drift based on userdata and other ec2 level settings.

This is a simple way to make CLM behave similar and be responsible for the node rotation triggering while other tools like CLC will still work with this. In the future it makes sense to look into the native drift feature in Karpenter to avoid implementing half of it in CLM.

### Depends on

* [x] #949

### DIFF
--- a/pkg/updatestrategy/aws_ec2.go
+++ b/pkg/updatestrategy/aws_ec2.go
@@ -76,7 +76,12 @@ func NewEC2NodePoolBackend(cluster *api.Cluster, cfg aws.Config, karpenterClient
 // userData,ImageID and tags and 'outdated' for nodes with an outdated
 // configuration.
 func (n *EC2NodePoolBackend) Get(ctx context.Context, nodePool *api.NodePool) (*NodePool, error) {
-	if nodePool.Profile == "worker-karpenter" {
+	// scope the Karpenter Drift detection logic to only Karpenter node
+	// pools that use Bottlerocket or AL2023 AMIs.
+	// This is done to limit the change to the initial roll out as Drift
+	// detection has a potential issue as outlined in:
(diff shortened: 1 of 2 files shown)


mikkeloscar:
I observed nodeclaims drift when it wasn't expected which would cause this change to rotate nodes much more often.

Suspected issue on Karpenter side: https://github.com/aws/karpenter-provider-aws/pull/9083

mikkeloscar:
> I observed nodeclaims drift when it wasn't expected which would cause this change to rotate nodes much more often.
> 
> Suspected issue on Karpenter side: [aws/karpenter-provider-aws#9083](https://github.com/aws/karpenter-provider-aws/pull/9083)

This was changed such that the new drift detection logic is only in effect when using EKS AMIs, this allows us to roll this out more granular without affecting legacy AMI behavior. We still need to understand how to deal with this issue before being able to roll out EKS AMIs widely.

mikkeloscar:
:+1:

zaklawrencea (approved):

PR #954, opened 2026-08-03
Disable coverage report

mikkeloscar:
Code Quality is disabled and can't be enabled at repo level. Need to clarify internally if it can be enabled or an alternative is needed.

Until then we disable it to avoid broken builds.


mikkeloscar:
:+1:

zaklawrencea (approved):

PR #961, opened 2026-08-26
feat: deletions skip kind not exists to cover deleted CRDs

tcondeixa:
Root cause: ResolveKind was silently discarding mapper errors (using _), then generating its own opaque error string when the GVR was empty. Callers couldn't detect whether the failure was a NoMatchError/AmbiguousError (CRD not registered) vs a genuinely misconfigured kind — so deleteIfFound couldn't treat it as "already gone, skip gracefully."

Changes in pkg/kubernetes/kubernetes.go:

1. resolveKind (new private helper) — contains the original ResolveKind logic but now preserves the real mapper error instead of discarding it with _. The first error is kept if both lookups fail, so the more specific one is always returned. Falls back to the original error message only when the mapper returns an empty GVR with no error.
2. ResolveKind — now wraps resolveKind with a reset-and-retry on NoMatchError or AmbiguousError, matching the pattern already used in getRESTMappingFor (kubectl_apply.go). This handles stale discovery cache without permanently failing.
3. deleteIfFound — now skips gracefully on IsNoMatchError or IsAmbiguousError instead of hard-failing. Covers both NoKindMatchError (CRD deleted) and AmbiguousResourceError (CRD has multiple versions registered).
4. DeleteResource — same graceful skip applied to the ListResources call in the selector/labels path.

NOTE: Generate with help of AI

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -191,15 +191,27 @@ func (c *ClientsCollection) getResourceClient(kind, namespace string) (dynamic.R
 
 func (c *ClientsCollection) ResolveKind(kind string) (schema.GroupVersionResource, error) {
 	var gvr schema.GroupVersionResource
+	var lastErr error
 	fullySpecifiedGVR, groupResource := schema.ParseResourceArg(kind)
 
 	if fullySpecifiedGVR != nil {
-		gvr, _ = c.Mapper.ResourceFor(*fullySpecifiedGVR)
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

tcondeixa:
After a review I improved it a bit, can you check again @mikkeloscar? thanks

tcondeixa:
👍

mikkeloscar:
👍

PR #966, opened 2026-09-18
Add seed-default profile

zaklawrencea:
Add a new `seed-default` profile which enables the provisioning of EKS seed nodes utilising Bottlerocket. The profile configuration is defined in the [kubernetes-on-aws](https://github.com/zalando-incubator/kubernetes-on-aws) repo and introduced in: https://github.com/zalando-incubator/kubernetes-on-aws/pull/12292


zaklawrencea:
Changes no longer required. Seed nodes will be moved to Bottlerocket without the use of a new profile.

